PR #21, opened 2022-01-28
Course apis

e-wai:
## Implementation Description
Brief summary of PR

## Screenshots
Visual depiction of PR

- 

## Testing Procedure

## Things to Note

## Checklist
- [ ] My code follows the style guide


backend/graphql/types/courseType.ts
```diff
@@ -0,0 +1,33 @@
+import { gql } from "apollo-server-express";
+
+const courseType = gql`
+  type CourseResponseDTO {
+    id: ID!
+    title: String!
+    description: String
+    image: String
+    previewImage: String
+    lessons: [ID!]!
```
draymond63, line 10: Just so I understand, does this mean there has to be a lesson ID within the list? Can the list be be empty?

backend/models/course.model.ts
```diff
@@ -0,0 +1,32 @@
...
+  image: {
+    type: String,
+  },
+  previewImage: {
+    type: String,
+  },
+  lessons: {
+    type: [String],
+    required: true,
+  },
```
draymond63, line 29: Isn't the supposed to be a list of objects now?

backend/graphql/types/courseType.ts
e-wai: From the docs (https://graphql.org/learn/schema/):
> [Episode!]! represents an array of Episode objects. Since it is also non-nullable, you can always expect an array (with zero or more items) when you query the appearsIn field. And since Episode! is also non-nullable, you can always expect every item of the array to be an Episode object.

Sounds like it guarantees it has to be an array always, and content of the array has to be that type always

backend/models/course.model.ts
e-wai: Isn't that modules... I think?

backend/graphql/types/courseType.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  input CourseRequestDTO {
+    title: String!
+    description: String
+    image: String
+    previewImage: String
+    lessons: [ID!]!
```
Shrinjay, line 18: lessons here would need to be a list of Lesson objects as well

backend/middlewares/validators/courseValidators.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  if (!validatePrimitive(body.image, "string")) {
+    return res.status(400).send(getApiValidationError("image", "string"));
+  }
+  if (!validatePrimitive(body.previewImage, "string")) {
+    return res
+      .status(400)
+      .send(getApiValidationError("previewImage", "string"));
+  }
+  if (!validateArray(body.lessons, "string")) {
```
Shrinjay, line 36: Validation would probably need to be updated to an array of objects for lessons, doesn't look like util.ts has a validator for custom objects though so maybe we can just drop validation on lessons or write our own validation function.

backend/services/interfaces/ICourseService.ts
```diff
@@ -0,0 +1,63 @@
+export interface CourseRequestDTO {
+  id: string;
+  title: string;
+  description: string;
+  image: string;
+  previewImage: string;
+  lessons: [string];
```
Shrinjay, line 7: DTOs also need to be updated for lessons

Shrinjay (commented):
Left a few comments around the lessons being array of objects vs. array of strings discussion. Based off what Daniel said at the standup and the discussion last work session, it's going to be an array of objects, so code should probably reflect that, but we can have more discussions around this if you'd like.

Since we don't need to actually implement anything other than the lesson types (no endpoints or DTOs for lessons yet), should be a quick fix. Otherwise, this is super clean code, and it reminded me to go add validators to my PR so thanks LOL :)

backend/graphql/types/courseType.ts
draymond63: Whoops, I got confused @Shrinjay. I thought this was the lesson API PR. `lessons` will indeed be a list of IDs. Sorry for the confusion!

backend/models/course.model.ts
draymond63: Lol, this is totally my bad. We can discuss this further, but for now the current implementation is good. Again, sorry for the confusion everyone

draymond63 (approved):
Giving tentative approval; I'll let other's look over the code, but if y'all don't think it needs changing then I think it's good to go. Feel free to merge :)

backend/services/implementations/courseService.ts
```diff
@@ -0,0 +1,121 @@
...
+      title: course.title,
+      description: course.description,
+      image: course.image,
+      previewImage: course.previewImage,
+      lessons: course.lessons,
+    };
+  }
+
+  async getCourses(): Promise<CourseResponseDTO[]> {
+    try {
```
Shrinjay, line 36: Linter is going to yell at you for not using "this" in a class method. Technically, these methods should be static methods, but really doesn't matter. Just add `/* eslint-disable class-methods-use-this */` on the line above the getCourses and createCourse functions to stop the linter yelling at you, or use static methods, up to you.

Shrinjay (approved):
Sorry for completely getting the first review wrong due to a miscommunication, code looks super clean to me otherwise.

backend/services/implementations/courseService.ts
e-wai: Probably should keep it consistent in the repo. Thoughts on changing everything to use `this` vs just bypassing the linter?

e-wai:
Flagging up some discussion from slack:

> For course APIs, we are using the same request DTO for create and update, but fields that are mandatory for a create may not be for update

> If create and update have different required fields, I'd say make them separate DTOs

> Can use something like https://www.typescriptlang.org/docs/handbook/utility-types.html for cleanliness

backend/services/implementations/courseService.ts
```diff
@@ -0,0 +1,121 @@
...
+      }
+    } catch (error: unknown) {
+      Logger.error(
+        `Failed to update course. Reason = ${getErrorMessage(error)}`,
+      );
+      throw error;
+    }
+
+    return {
+      id: course.id,
```
e-wai, line 96: will return null, move `updatedCourse` declaration and change returns
```suggestion
      id: updatedCourse.id,
```

backend/services/implementations/courseService.ts
Shrinjay: I don't really see how you'd use `this` inside the method? There's no state being kept in the service itself. The pattern in the repo seems to be to bypass the linter in this case, look at the authService.ts file for reference.

backend/services/implementations/courseService.ts
```diff
@@ -0,0 +1,121 @@
...
+        previewImage: course.previewImage,
+        lessons: course.lessons,
+      }));
+    } catch (error: unknown) {
+      Logger.error(`Failed to get courses. Reason = ${getErrorMessage(error)}`);
+      throw error;
+    }
+  }
+
+  async createCourse(course: CourseRequestDTO): Promise<CourseResponseDTO> {
```
e-wai, line 52: Check if `create` operation returns null if it fails (aka `Promise<CourseRequestDTO>` vs. `Promise<CourseRequestDTO | null>`)

backend/services/implementations/courseService.ts
draymond63: I think changing it to `this` might be a bit cleaner instead of disabling the linter, but I can do that when I set up the linter this week. Doesn't matter too much for the PR

backend/services/implementations/courseService.ts
```diff
@@ -0,0 +1,121 @@
...
+      course = await MgCourse.findById(id);
+      if (!course) {
+        throw new Error(`Course id ${id} not found`);
+      }
+    } catch (error: unknown) {
+      Logger.error(`Failed to get course. Reason = ${getErrorMessage(error)}`);
+      throw error;
+    }
+
+    return {
```
e-wai, line 25: double check if need to split this or can do `return course`?

backend/services/implementations/courseService.ts
draymond63: Yeah I'm not a big fan of how starter-code implements it. Like, `newCourse` will always be a Course type, never null. If it is null, it's because an error was thrown and it will never be returned anyway. My vote would be to return null if it fails

backend/services/implementations/courseService.ts
draymond63: `MgCourse.findById` returns `Course`,  not `CourseResponseDTO`. Not sure if that makes a big difference: it seems like typescript can see that they both have the same attributes. I'd save leave it for now, but it's something we can discuss at the meeting on Thursday

draymond63: Perhaps even something like `course as CourseResponseDTO` would be good

backend/services/implementations/courseService.ts
e-wai: As in, `newCourse` will never be null even when it fails because it'll throw an error?

PR #22, opened 2022-01-28
Correct user signup input

Shrinjay:
## Implementation Description
Adds a town field to the user signup page

## Screenshots
image

Correctness of the data model on both create, update and read operations for users

## Testing Procedure

- Test backend to ensure that api calls for users and authentication for create and read return a town property
- Test frontend to ensure that town property is in form and is updated correctly when input is changed
- E2E test frontend and backend to ensure that a user can be created with a given town name and the town name is returned when the user is queried for

Completed tests:
- As part of this ticket, I ran 2 quick tests locally. 1) Signup with a user, ensure that town property is in API request and saved in database. 2) Login with user, ensure that town property is returned as part of the model. Both tests passed. 
- Note that I didn't try and run a user query test with graphql playground because it requires me to have an admin login which I don't have. 

Points of failure include: Form input, API request to create user, API response on user query, saving user to database, reading user from database.

## Things to Note
- As part of this PR, I've also removed the google signin logic as we've determined that it is not a requirement for this project and it makes the code for this issue a little more complicated.

### DIFF
--- a/backend/graphql/resolvers/authResolvers.ts
+++ b/backend/graphql/resolvers/authResolvers.ts
@@ -31,16 +31,6 @@ const authResolvers = {
       res.cookie("refreshToken", refreshToken, cookieOptions);
       return rest;
     },
-    loginWithGoogle: async (
-      _parent: undefined,
-      { idToken }: { idToken: string },
-      { res }: { res: Response },
-    ): Promise<Omit<AuthDTO, "refreshToken">> => {
-      const authDTO = await authService.generateTokenOAuth(idToken);
-      const { refreshToken, ...rest } = authDTO;
-      res.cookie("refreshToken", refreshToken, cookieOptions);
-      return rest;
-    },
     register: async (
       _parent: undefined,
       { user }: { user: RegisterUserDTO },
--- a/backend/services/implementations/authService.ts
+++ b/backend/services/implementations/authService.ts
@@ -38,45 +38,6 @@ class AuthService implements IAuthService {
     }
   }
 
-  /* eslint-disable class-methods-use-this */
-  async generateTokenOAuth(idToken: string): Promise<AuthDTO> {
-    try {
-      const googleUser = await FirebaseRestClient.signInWithGoogleOAuth(
-        idToken,
-      );
-      // googleUser.idToken refers to the Firebase Auth access token for the user
-      const token = {
-        accessToken: googleUser.idToken,
-        refreshToken: googleUser.refreshToken,
-      };
-      // If user already has a login with this email, just return the token
-      try {
-        // Note: an error message will be logged from UserService if this lookup fails.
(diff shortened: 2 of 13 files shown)


draymond63 (changes requested):
Hey Shrinjay! Most of the real edits look good from what I can tell so far. 

The main issue is there are a lot of files where git says the entire thing is being edited. If you look at the "Files changed" in the PR, it is pretty difficult to tell what the real changes are. This can cause problems for two reasons:

1. It makes the review more difficult. There's a lot of identical code the Git says is new

2. It makes is much more difficult for others to merge their code in, because Git will think it conflicts with your changes.

So, please reset your commits and be careful about adding files that shouldn't be (avoid `git add .`). Pay attention to **which files** are noted as changed and what in those files are being changed. VS code is very helpful for this.

If you have any questions about git, please let me know!
image

Shrinjay:
Hey Daniel, sorry about that! I didn't notice a ton of these files got added, I'll fix it quickly.

Shrinjay:
> Hey Daniel, sorry about that! I didn't notice a ton of these files got added, I'll fix it quickly.

HAH it was the linter, I'm not quite sure what's going on with the linter, seems it tried to fix a bunch of "issues" but it didn't change anything, and the code still failed linting. Seems our deployment workflow is failing as well, not sure what's going on with it all but yeah.

backend/services/interfaces/authService.ts
```diff
@@ -11,15 +11,6 @@ interface IAuthService {
    */
   generateToken(email: string, password: string): Promise<AuthDTO>;
 
-  /**
```
Shrinjay, line 14: Removed since we also wanted to get rid of Google OAuth here since it won't be used and messes up some of the changes.

backend/services/implementations/authService.ts
```diff
@@ -38,45 +38,6 @@ class AuthService implements IAuthService {
     }
   }
 
-  /* eslint-disable class-methods-use-this */
```
Shrinjay, line 41: Removed since we won't use Google OAuth

backend/graphql/resolvers/authResolvers.ts
```diff
@@ -31,16 +31,6 @@ const authResolvers = {
       res.cookie("refreshToken", refreshToken, cookieOptions);
       return rest;
     },
-    loginWithGoogle: async (
```
Shrinjay, line 34: Removed since we won't use Google OAuth and having it messes up some of the changes in this ticket.

draymond63 (approved):
The large git diff would *normally* be a no go, but seeing as this is due to my fault with linting issues and they shouldn't conflict with anybody else's PR, I'm okay with. I'm going to let a few other people review it to see if they catch anything, but other then that I think it's good to go.

PR #34, opened 2022-01-30
Implement lesson create and read APIs

y2kwak:
## Implementation Description
Created backend `create` and `read` APIs for lessons

## Screenshots
Mutation:
Screen Shot 2022-01-27 at 22 00 56

Query:
Screen Shot 2022-01-30 at 15 14 31

- Does this reflect the data models outlined on the notion [page](https://www.notion.so/uwblueprintexecs/Data-Models-f6b8178406c1481da654a70e30eabe0e#a41f569f852f4d2e9d569990f67e1563)?
- Are the correct fields marked optional?

## Testing Procedure

Testing procedure:
- Go to <backend-url>/graphql, for me this is [http://localhost:5000/graphql](http://localhost:5000/graphql)
- Try out queries and mutations. Include some mutations that don't have default values included (eg. no description):
Mutation 
```
mutation{
createLesson(lesson: {
    course: "ID",
    title:"Lesson 1",
    content: [],
  }) {
    id,
    course,
    title,
    description,
    image,
    content,
  }
}
```

Query (single lesson by id)
```
query {
  lessonById(id : <insert lesson id>
  ) {
    id,
    course,
    title,
    description,
    image,
    content,
  }
}
```

## Things to Note

## Checklist
- [ ] My code follows the style guide


backend/middlewares/validators/lessonValidators.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  getApiValidationError,
+  validateArray,
+  validatePrimitive,
+} from "./util";
+import { getErrorMessage } from "../../utilities/errorUtils";
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable-next-line import/prefer-default-export */
+export const courseRequestDtoValidator = async (
```
e-wai, line 11: `lessonRequestDtoValidator`? Also I'm not sure you have a `LessonRequestDTO` right now, only `LessonDTO`

backend/middlewares/validators/lessonValidators.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!validatePrimitive(body.image, "string")) {
+    return res.status(400).send(getApiValidationError("image", "string"));
+  }
+  if (!validateArray(body.content, "string")) {
+    return res
+      .status(400)
+      .send(getApiValidationError("content", "string"));
+  }
+  return next();
+};
```
e-wai, line 40: newline unless it tells you not to :)

backend/models/lesson.model.ts
```diff
@@ -0,0 +1,36 @@
...
+    type: String,
+    required: true,
+  },
+  title: {
+    type: String,
+    required: true,
+  },
+  description: {
+    type: String,
+    required: false,
```
e-wai, line 23: Hmm consistency thing - is there a preference between not including `required` attribute vs. specifying `required: false`?

backend/services/implementations/lessonService.ts
```diff
@@ -0,0 +1,61 @@
...
+  /* eslint-disable class-methods-use-this */
+  async getLessonById(id: string): Promise<LessonDTO> {
+    let lesson: Lesson | null;
+    try {
+      lesson = await MgLesson.findById(id);
+      if (!lesson) {
+        throw new Error(`Lesson id ${id} not found`);
+      }
+    } catch (error: unknown) {
+      Logger.error(`Failed to get entity. Reason = ${getErrorMessage(error)}`);
```
e-wai, line 25:
```suggestion
      Logger.error(`Failed to get lesson. Reason = ${getErrorMessage(error)}`);
```

backend/services/interfaces/lessonService.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * create an Lesson with the fields given in the DTO, return created Lesson
+   * @param lesson to be created
+   * @returns the created Lesson
+   * @throws Error if creation fails
+   */
+  createLesson(lesson: CreateLessonDTO): Promise<LessonDTO>;
+}
+
```
e-wai, line 23: EOF seems weird

backend/models/lesson.model.ts
```diff
@@ -0,0 +1,36 @@
...
+  },
+  image: {
+    type: String,
+    required: false,
+  },
+  content: {
+    type: [Object],
```
Shrinjay, line 30: I don't think `Object` is the right way to define this in Mongoose? Mongoose has certain [SchemaTypes](https://mongoosejs.com/docs/schematypes.html#mixed) and I think the Mixed SchemaType is best for representing an object we haven't defined yet.

draymond63 (approved):
Things I believe look good, giving a tentative approval (just waiting to see if others have any thoughts)

backend/models/lesson.model.ts
```diff
@@ -0,0 +1,36 @@
+import mongoose, { Schema, Document } from "mongoose";
+
```
Shrinjay, line 2: Maybe add a TODO here that says "Replace Object type with Content type"? It'll be helpful once we get those content types defined to find where we need to update.

backend/graphql/types/lessonType.ts
```diff
@@ -0,0 +1,32 @@
...
+// TO DO: Fix content typing
+
+const lessonType = gql`
+  type LessonDTO {
+    id: ID!
+    course: ID!
+    title: String
+    description: String
+    image: String
+    content: [ID!]
```
Shrinjay, line 12: content should be an array of Content objects not IDs (see Slack for that discussion). I don't think there's something like the `Object` type in GraphQL, but if you can find one, feel free to use it. Otherwise, just define a dummy `Content` GraphQL type and put it in here. It can just be an empty type for now.

backend/graphql/types/lessonType.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  input CreateLessonDTO {
+    course: ID!
+    title: String
+    description: String
+    image: String
+    content: [ID!]
```
Shrinjay, line 20: Similarly, content should be a list of objects

backend/middlewares/validators/lessonValidators.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!validatePrimitive(body.title, "string")) {
+    return res.status(400).send(getApiValidationError("title", "string"));
+  }
+  if (!validatePrimitive(body.description, "string")) {
+    return res.status(400).send(getApiValidationError("description", "string"));
+  }
+  if (!validatePrimitive(body.image, "string")) {
+    return res.status(400).send(getApiValidationError("image", "string"));
+  }
+  if (!validateArray(body.content, "string")) {
```
Shrinjay, line 34: Replacing content with list of objects will also affect the validator, we can either write our own validator for the type or just not validate it, up to you.

backend/services/implementations/lessonService.ts
```diff
@@ -0,0 +1,61 @@
...
+      throw error;
+    }
+
+    return {
```
Shrinjay, line 29: I'm honestly not sure if there's any benefit in defining a new object with `return {...` to copy over all the properties of lesson instead of just doing `return lesson`? Not a huge thing.

backend/graphql/index.ts
```diff
@@ -42,6 +44,7 @@ const graphQLMiddlewares = {
     userById: authorizedByAdmin(),
     userByEmail: authorizedByAdmin(),
     users: authorizedByAdmin(),
+    lessonById: authorizedByAllRoles(),
```
e-wai, line 47: `lessonById` seems redundant since there's only one lesson query

backend/services/implementations/lessonService.ts
```diff
@@ -0,0 +1,61 @@
...
+      description: lesson.description,
+      image: lesson.image,
+      content: lesson.content,
+    };
+  }
+
+  async createLesson(lesson: CreateLessonDTO): Promise<LessonDTO> {
+    let newLesson: Lesson;
+    try {
+      newLesson = await MgLesson.create({ ...lesson });
```
Shrinjay, line 42: Why are we unpacking the lesson argument into a new object with `{...lesson}`. Afaik this is just the same as going `MgLesson.create(lesson)`?

backend/types.ts
```diff
@@ -22,6 +22,17 @@ export type RegisterUserDTO = Omit<CreateUserDTO, "role">;
...
+export type LessonDTO = {
+  id: string;
+  course: string;
+  title: string;
+  description: string;
+  image: string;
+  content: [Object];
+}
+
+export type CreateLessonDTO = Omit<LessonDTO, "id">;
```
Shrinjay, line 34: For consistency, this should probably be `CreateLessonRequestDTO`, given that it's a request, not an entity.

backend/graphql/types/lessonType.ts
```diff
@@ -0,0 +1,32 @@
...
+  type LessonDTO {
+    id: ID!
+    course: ID!
+    title: String
+    description: String
+    image: String
+    content: [ID!]
+  }
+
+  input CreateLessonDTO {
```
Shrinjay, line 15: Like below, we should maybe this CreateLessonRequestDTO instead?

backend/types.ts
```diff
@@ -22,6 +22,17 @@ export type RegisterUserDTO = Omit<CreateUserDTO, "role">;
 
 export type AuthDTO = Token & UserDTO;
 
+export type LessonDTO = {
```
e-wai, line 25: As a consistency thing (since our course APIs implementation should look pretty similar): we followed the `Entity` example more closely and declared the `CourseRequestDTO`/`CourseResponseDTO` as interfaces in the `services/interfaces/ICourseService`.
- thoughts on interfaces vs. types for DTO declarations?
- calling classes in `services/interfaces` as `ILessonService` vs. `lessonService`
- might make sense to specify that this is the `LessonResponseDTO`

backend/services/implementations/lessonService.ts
e-wai, line 42: will this ever return a null type?

Shrinjay (changes requested):
Content needs to be changed from array of Ids to array of objects in the graphql type, couple other naming, formatting nitpicks from me. Really well done code otherwise :)

e-wai (changes requested):
Some last few comments, I think it looks pretty good for the most part - some of my comments are just making sure that we (course API people) are staying consistent with you guys

backend/graphql/types/authType.ts
```diff
@@ -6,6 +6,7 @@ const authType = gql`
     firstName: String!
     lastName: String!
     email: String!
+    town: String!
```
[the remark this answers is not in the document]
draymond63: This should be fixed.

backend/graphql/types/lessonType.ts
draymond63: +1 to this

PR #35, opened 2022-01-30
Remove Google Auth

Shrinjay:
## Implementation Description
Since we already removed some parts of google auth in a past PR, Emma suggested I delete it from the frontend auth service, at which point I figured I'd just nuke Google Auth from the codebase. Perhaps it's because I hold a grudge from them rejecting my application, but it saves us work down the line. The google auth code still exists in start code in case we want to reuse the design pattern.

## Screenshots
Login screen with no google auth:
image

- That this didn't break anything else in auth

## Testing Procedure
- Try logging in, logging out, signup, other auth operations. 
- Completed tests: I've already tried singing up, then logging in, then logging out, seems to work fine.

### DIFF
--- a/backend/graphql/types/authType.ts
+++ b/backend/graphql/types/authType.ts
@@ -21,7 +21,6 @@ const authType = gql`
 
   extend type Mutation {
     login(email: String!, password: String!): AuthDTO!
-    loginWithGoogle(idToken: String!): AuthDTO!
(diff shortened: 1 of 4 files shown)


draymond63 (approved):
Looks good! I'll merge it in. Thanks for doing that Shrinjay

PR #42, opened 2022-02-04
Implement Lesson API - Create and Delete

y2kwak:
## Implementation Description
Added create and delete mutations for lessons!

## Screenshots
N/A Backend work

- Does this align with how we want to add new APIs? Are we relying properly on the entity example?
- Are the correct fields marked optional?

## Testing Procedure
Testing procedure:
- Go to <backend-url>/graphql, for me this is [http://localhost:5000/graphql](http://localhost:5000/graphql)
- Try out queries and mutations. Include some mutations that don't have default values included (eg. no description):
Mutation 
```
mutation{
// create a lesson
createLesson(lesson: {
    course: "ID",
    title:"Lesson 1",
    content: [""],
  }) {
    id,
    course,
    title,
    description,
    image,
    content,
  }
}
// change a parameter
updateLesson(id: string, lesson: {
    course: "ID",
    description:"this is a modified description!",
    content: [""],
  }) {
    id,
    course,
    title,
    description,
    image,
    content,
  }
}
// delete a lesson, and query to check if exists
deleteLesson(id: string) : {
  id
}
```

## Things to Note
N/A

## Checklist

### DIFF
--- a/backend/graphql/resolvers/lessonResolvers.ts
+++ b/backend/graphql/resolvers/lessonResolvers.ts
@@ -2,7 +2,8 @@ import LessonService from "../../services/implementations/lessonService";
 import {
   ILessonService,
   LessonResponseDTO,
-  LessonRequestDTO,
+  LessonCreateRequestDTO,
+  LessonUpdateRequestDTO,
 } from "../../services/interfaces/ILessonService";
 
 const lessonService: ILessonService = new LessonService();
@@ -19,14 +20,14 @@ const lessonResolvers = {
   Mutation: {
     createLesson: async (
       _parent: undefined,
-      { lesson }: { lesson: LessonRequestDTO },
+      { lesson }: { lesson: LessonCreateRequestDTO },
     ): Promise<LessonResponseDTO> => {
       const newLesson = await lessonService.createLesson(lesson);
       return newLesson;
     },
     updateLesson: async (
       _parent: undefined,
-      { id, lesson }: { id: string; lesson: LessonRequestDTO },
+      { id, lesson }: { id: string; lesson: LessonUpdateRequestDTO },
     ): Promise<LessonResponseDTO | null> => {
       return lessonService.updateLesson(id, lesson);
     },
--- a/backend/graphql/resolvers/lessonResolvers.ts
+++ b/backend/graphql/resolvers/lessonResolvers.ts
@@ -24,6 +24,18 @@ const lessonResolvers = {
       const newLesson = await lessonService.createLesson(lesson);
       return newLesson;
     },
+    updateLesson: async (
+      _parent: undefined,
+      { id, lesson }: { id: string; lesson: LessonRequestDTO },
+    ): Promise<LessonResponseDTO | null> => {
+      return lessonService.updateLesson(id, lesson);
+    },
+    deleteLesson: async (
+      _parent: undefined,
+      { id }: { id: string },
+    ): Promise<string> => {
+      return lessonService.deleteLesson(id);
+    },
   },
 };
 export default lessonResolvers;
--- a/backend/graphql/resolvers/lessonResolvers.ts
+++ b/backend/graphql/resolvers/lessonResolvers.ts
@@ -2,8 +2,8 @@ import LessonService from "../../services/implementations/lessonService";
 import {
   ILessonService,
   LessonResponseDTO,
-  LessonCreateRequestDTO,
-  LessonUpdateRequestDTO,
+  CreateLessonRequestDTO,
+  UpdateLessonRequestDTO,
 } from "../../services/interfaces/ILessonService";
 
 const lessonService: ILessonService = new LessonService();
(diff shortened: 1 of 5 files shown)


draymond63:
Seems like the linter failed - run the linter using the commands found in the README, resolve any of the errors/warnings, and commit those changes!

backend/graphql/resolvers/lessonResolvers.ts
```diff
@@ -24,6 +24,18 @@ const lessonResolvers = {
       const newLesson = await lessonService.createLesson(lesson);
       return newLesson;
     },
+    updateLesson: async(
+      _parent: undefined,
+      { id, lesson }: { id: string, lesson: LessonRequestDTO },
+    ): Promise<LessonResponseDTO> => {
```
e-wai, line 30: I swear that update could fail with a `null` return... maybe check Entity?

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
e-wai, line 1: Was this intentionally changed?

e-wai (changes requested):
Main comment is on the `package-lock.json` change, but everything looks good so far
-------
When we were manually testing, I believe that the model object required/validated that all fields marked as mandatory on the model were included in the CreateLesson, so then the UpdateLesson call can be made with optional params dictated by the DTO object.

I would suggest that the DTO object should actually be
```
type LessonResponseDTO {
    id: ID!
    course: ID
    title: String
    description: String
    image: String
    content: [Content!]!
  }
```
since we won't always want to update course and content ! just means that at least an empty array is guaranteed, guaranteed to only be Content objects

@draymond63 @Shrinjay think we were talking about Create vs Update DTO types; if model handles validation, could we combine them together still? Or keep them distinct for clarity

backend/graphql/types/lessonType.ts
```diff
@@ -28,6 +28,8 @@ const lessonType = gql`
 
   extend type Mutation {
     createLesson(lesson: LessonRequestDTO!): LessonResponseDTO!
+    updateLesson(id: ID!, lesson: LessonRequestDTO!): LessonResponseDTO!
```
Shrinjay, line 31: Regardless of if create and update use the same DTO or different ones, I think we need a more clear name than `LessonRequestDTO`, it comes off as a DTO to get a lesson. Something like `LessonCreateRequestDTO` or `LessonCreateUpdateRequestDTO` (if we choose to use the same DTO` would be more appropriate imo.

Shrinjay (commented):
Minor variable naming nitpick but it's not a huge deal, but let's resolve the one v. two DTO question for create and update

backend/services/interfaces/ILessonService.ts
```diff
@@ -31,4 +31,24 @@ export interface ILessonService {
...
+   * @returns the updated Lesson
+   * @throws Error if update fails
+   */
+  updateLesson(
+    id: string,
+    lesson: LessonRequestDTO,
+  ): Promise<LessonResponseDTO | null>;
+
+  /**
+   * create an Lesson with the fields given in the DTO, return created Lesson
```
draymond63, line 48: create "a" lesson

backend/services/interfaces/userService.ts
```diff
@@ -1,10 +1,4 @@
-import {
-  CreateUserDTO,
-  Role,
-  SignUpMethod,
-  UpdateUserDTO,
-  UserDTO,
-} from "../../types";
+import { CreateUserDTO, Role, UpdateUserDTO, UserDTO } from "../../types";
```
draymond63, line 1: Why was the `SignUpMethod` removed in this PR? What did it do?

draymond63 (changes requested):
Please address comments - all very minor :)

Also, please remove the `package-lock.json` from the commits, as I don't think it is necessary for this PR. You will have to rewrite history which involves force pushing!

backend/services/interfaces/userService.ts
y2kwak: Oh, I think this changed when running the linter. I can remove it since [this PR ](https://github.com/uwblueprint/rowan-house/pull/48)seems to be handling it.

y2kwak: This will also mean that this PR won't pass linter until the other linter PR is merged and so on

draymond63 (approved):
This looks good to me! I'll wait to see if any of the other devs have any thoughts, but other than that I will merge it!

Shrinjay (approved):

e-wai (commented):
Nothing big, just a couple of small things 😀

backend/services/interfaces/ILessonService.ts
```diff
@@ -25,10 +33,30 @@ export interface ILessonService {
   getLessonById(id: string): Promise<LessonResponseDTO>;
```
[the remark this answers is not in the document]
y2kwak: I think this may be related to your previous comment [here](https://github.com/uwblueprint/rowan-house/pull/34#discussion_r795249898). But I think the same explanation applies here that if in the future we were to add more get Lesson methods, the specificity here would be beneficial.

PR #46, opened 2022-02-05
Fix updatedCourse mutation return bug where id not on argument

e-wai:
## Implementation Description
Found a bug on the `UpdateCourse` mutation -> calls `id` property of passed course, which doesn't exist. Get from result returned from db instead
- Change from prev PR, this comment -> https://github.com/uwblueprint/rowan-house/pull/21#discussion_r795252898

## Screenshots
Visual depiction of PR

- Makes sense? 

## Testing Procedure

See testing section in https://github.com/uwblueprint/rowan-house/pull/21

## Things to Note
- Additional dependencies/libraries you've added? None
- Things you'd want to know when coming back to the code after a few months Nothing new

## Checklist
- [x] I have linked the relevant issue in this PR -> just reference other PR

### DIFF
--- a/backend/services/implementations/courseService.ts
+++ b/backend/services/implementations/courseService.ts
@@ -74,15 +74,12 @@ class CourseService implements ICourseService {
     id: string,
     course: CourseRequestDTO,
   ): Promise<CourseResponseDTO | null> {
+    let updatedCourse: Course | null;
     try {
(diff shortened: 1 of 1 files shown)


Shrinjay (approved):
Makes sense to me, and I guess that's the acceptance criteria

y2kwak (approved):

PR #50, opened 2022-02-15
Adding private and public fields

vyomeshiyengar:
## Implementation Description
- Updated document models & all course APIs to include the published and private attribute(s)
- Added a 3rd role called STAFF to the list of possible roles users can have

## Screenshots
Visual depiction of PR

- Making sure the fields are added in the necessary locations, and that all the dependencies are met 

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/backend/graphql/types/courseType.ts
+++ b/backend/graphql/types/courseType.ts
@@ -18,8 +18,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
-    private: Boolean!
-    published: Boolean!
+    private: Boolean
+    published: Boolean
   }
 
   extend type Query {
--- a/backend/graphql/types/courseType.ts
+++ b/backend/graphql/types/courseType.ts
@@ -8,6 +8,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
+    private: Boolean
+    public: Boolean
   }
 
   input CourseRequestDTO {
@@ -16,6 +18,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
+    private: Boolean
+    public: Boolean
   }
 
   extend type Query {
--- a/backend/graphql/types/courseType.ts
+++ b/backend/graphql/types/courseType.ts
@@ -8,8 +8,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
-    private: Boolean
-    published: Boolean
+    private: Boolean!
+    published: Boolean!
   }
 
   input CourseRequestDTO {
@@ -18,8 +18,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
-    private: Boolean
-    published: Boolean
+    private: Boolean!
+    published: Boolean!
   }
 
   extend type Query {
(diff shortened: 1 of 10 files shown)


backend/graphql/types/courseType.ts
```diff
@@ -8,6 +8,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
+    private: Boolean
+    public: Boolean
```
draymond63, line 12: It would be good to call this property `published` as opposed to public.

backend/models/course.model.ts
```diff
@@ -27,6 +29,14 @@ const CourseSchema: Schema = new Schema({
     type: [String],
     required: true,
   },
+  private : {
+    type: Boolean,
+    default: false,
+  } ,
```
draymond63, line 35: Please run the linter and push a new commit with the changes it makes. If you have any concerns/questions, lmk

draymond63 (changes requested):
Sweet! I haven't tested if it works, but it looks good. There's just a couple changes I would make :)

backend/services/interfaces/ICourseService.ts
```diff
@@ -5,6 +5,8 @@ export interface CourseRequestDTO {
   image: string;
```
e-wai, line 5: Do we want people creating the course to have control over the private/public states when created/updated? I don't 100% remember the flow, but I'm thinking that when a course is created it defaults to not published.

cc: @draymond63

backend/models/course.model.ts
vyomeshiyengar: Done :)

backend/graphql/types/courseType.ts
vyomeshiyengar: Done

draymond63 (approved):
Looks good to me. I haven't tested to see if it works, but assuming it does I'm happy :)

backend/services/implementations/courseService.ts
```diff
@@ -97,6 +103,8 @@ class CourseService implements ICourseService {
       image: updatedCourse.image,
       previewImage: updatedCourse.previewImage,
       lessons: updatedCourse.lessons,
+      private: course.private,
+      published: course.published,
```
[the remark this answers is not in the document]
vyomeshiyengar: Should this be an optional return then or just not have this as a field there?

backend/graphql/types/courseType.ts
```diff
@@ -16,6 +18,8 @@ const courseType = gql`
     image: String
     previewImage: String
     lessons: [ID!]!
+    private: Boolean!
+    published: Boolean!
```
draymond63, line 22: I thought we decided that these shouldn't be mandatory - both defaulting to false? This is the request DTO, which is going to be sent from the frontend to the backend.

backend/services/implementations/courseService.ts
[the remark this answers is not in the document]
draymond63: I can't see where this is in the file, but I think we decided that the response DTO should have these fields as mandatory, but the request DTOs have them as optional, right?

Shrinjay (approved):

e-wai (approved):

PR #51, opened 2022-02-18
Implement aggregate user count by town query

e-wai:
## Implementation Description
Rowan house wants to be able to query information about their users -- one such query is number of users per town, limited by a start and end date
- Uses objectId timestamps, autogenerated instead of creating our own field. Docs here -> https://docs.mongodb.com/manual/reference/method/ObjectId.getTimestamp/
- Returns JSON string (`"{\"Toronto\":2,\"Vancouver\":1}"`) which is easily understood and can be parsed
  - However, this is not labelled as town / userCount so please comment on whether it's clear and usable
- Future improvement in frontend -- limit input selection to prevent typos, etc.

## Screenshots
Visual depiction of PR **n/a**

- Does return format make sense?
- Do we need further granularity than `yyyy/mm/dd`?
- No frontend work in this ticket?

## Testing Procedure

**Query endpoint**
(First, add users to the database. Hard to test right away since this relies on objectId timestamps -- which are autogenerated)
No parameters
```
{
  userCountByTown
}

```
Single date provided
- enddate: before all data, some data included, after all data, future
- startdate: before all data, some data included, after all data, future
```
{
  userCountByTown(endDate: "2022/01/01")
}
```
Both dates provided
- valid date rate (before < after)
- invalid (after < before)
- expect results within timerange
- do not expect result within timerange
- expect some results within time range
```
{
  userCountByTown(startDate: "2020/01/01", endDate: "2022/01/01")
}
```

## Things to Note
  - Did JSON string format work for frontend?

## Checklist

### DIFF
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -62,6 +62,7 @@ const graphQLMiddlewares = {
     userById: authorizedByAdmin(),
     userByEmail: authorizedByAdmin(),
     users: authorizedByAdmin(),
+    userCountByTown: authorizedByAdmin(),
     lessonById: authorizedByAllRoles(),
   },
   Mutation: {
--- a/backend/graphql/types/userType.ts
+++ b/backend/graphql/types/userType.ts
@@ -37,6 +37,7 @@ const userType = gql`
     userByEmail(email: String!): UserDTO!
     users: [UserDTO!]!
     usersCSV: String!
+    userCountByTown(startDate: String, endDate: String): String!
   }
 
   extend type Mutation {
--- a/backend/services/implementations/userService.ts
+++ b/backend/services/implementations/userService.ts
@@ -1,4 +1,5 @@
 import * as firebaseAdmin from "firebase-admin";
+import { Types } from "mongoose";
 
 import IUserService from "../interfaces/userService";
 import MgUser, { User } from "../../models/user.model";
@@ -15,6 +16,50 @@ const getMongoUserByAuthId = async (authId: string): Promise<User> => {
   }
   return user;
 };
+
+/**
+ * Utility method for converting yyyy/mm/dd date
+ *  string to 24-char hex timestamp string
+ *
+ * ObjectID docs:
+ *  https://mongodb.github.io/node-mongodb-native/api-bson-generated/objectid.html
+ * Reference:
+ *  https://kchodorow.com/2011/12/20/querying-for-timestamps-using-objectids/
+ */
+const getDateTime = (date: string): string =>
+  `${Math.floor(new Date(date).getTime() / 1000).toString(16)}0000000000000000`;
+
+interface LimitByDateQuery {
+  $match: {
+    _id: {
+      [key: string]: Types.ObjectId;
+    };
+  };
+}
+
+const buildLimitByDateQuery = (
+  startDate?: string,
+  endDate?: string,
+): LimitByDateQuery => {
+  const matchQuery: LimitByDateQuery = { $match: { _id: {} } };
+
+  if (startDate) {
+    // eslint-disable-next-line no-underscore-dangle
+    matchQuery.$match._id.$gt = Types.ObjectId(getDateTime(startDate));
+  }
+
+  if (endDate) {
+    // eslint-disable-next-line no-underscore-dangle
+    matchQuery.$match._id.$lt = Types.ObjectId(getDateTime(endDate));
+  }
+
+  return matchQuery;
+};
+
+interface UserCountGroupByTown {
+  _id: string;
+  userCount: number;
+}
 
 class UserService implements IUserService {
   /* eslint-disable class-methods-use-this */
@@ -141,6 +186,33 @@ class UserService implements IUserService {
     }
 
     return userDtos;
+  }
+
+  async getUserCountByTown(
+    startDate?: string,
+    endDate?: string,
+  ): Promise<{ [key: string]: number }> {
+    const groupByQuery = { $group: { _id: "$town", userCount: { $sum: 1 } } };
+
+    let query: Array<UserCountGroupByTown>;
+
+    if (!startDate && !endDate) {
+      query = await MgUser.aggregate([groupByQuery]).exec();
+    } else {
+      query = await MgUser.aggregate([
+        buildLimitByDateQuery(startDate, endDate),
+        groupByQuery,
+      ]).exec();
+    }
+
+    return query.reduce(
+      (map: { [key: string]: number }, obj: UserCountGroupByTown) => {
+        // eslint-disable-next-line no-param-reassign, no-underscore-dangle
+        map[obj._id] = obj.userCount;
+        return map;
+      },
+      {},
+    );
   }
 
   async createUser(user: CreateUserDTO): Promise<UserDTO> {
--- a/backend/graphql/resolvers/userResolvers.ts
+++ b/backend/graphql/resolvers/userResolvers.ts
@@ -34,6 +34,13 @@ const userResolvers = {
       const csv = await generateCSV<UserDTO>({ data: users });
       return csv;
     },
+    userCountByTown: async (
+      _parent: undefined,
+      { startDate, endDate }: { startDate?: string; endDate?: string },
+    ): Promise<string> => {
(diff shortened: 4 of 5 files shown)


backend/graphql/index.ts
```diff
@@ -62,6 +62,7 @@ const graphQLMiddlewares = {
     userById: authorizedByAdmin(),
     userByEmail: authorizedByAdmin(),
     users: authorizedByAdmin(),
+    userCountByTown: authorizedByAllRoles(),
```
e-wai, line 65: suspect this might be admin?

backend/services/implementations/userService.ts
```diff
@@ -143,6 +179,33 @@ class UserService implements IUserService {
...
+      query = await MgUser.aggregate([
+        buildMatchObject(startDate, endDate),
+        groupByQuery,
+      ]).exec();
+    }
+
+    return query.reduce(
+      (map: { [key: string]: number }, obj: UserCountGroupByTown) => {
+        // eslint-disable-next-line no-underscore-dangle
+        map[obj._id] = obj.userCount;
```
e-wai, line 202: Wasn't sure how to handle this lint error. (Is this actually indicative that something is wrong? I don't see the reassignment)

Can disable by changing line 201 to 
`// eslint-disable-next-line no-param-reassign, no-underscore-dangle`

backend/services/implementations/userService.ts
```diff
@@ -16,6 +17,41 @@ const getMongoUserByAuthId = async (authId: string): Promise<User> => {
...
+
+  if (startDate) {
+    // eslint-disable-next-line no-underscore-dangle
+    matchQuery.$match._id.$gt = Types.ObjectId(getDateTime(startDate));
+  }
+
+  if (endDate) {
+    // eslint-disable-next-line no-underscore-dangle
+    matchQuery.$match._id.$lt = Types.ObjectId(getDateTime(endDate));
+  }
```
e-wai, line 45: For a query with an open end (eg. look at all objects until now), is there any difference between saying `> startDate` vs `> startDate && <= Date.now`? in efficiency, style preference, etc.

backend/services/implementations/userService.ts
draymond63: Seems like this isn't a real error, and using that disable comment is fine. Seems like most people disable or even create a linter rule to ignore this error https://stackoverflow.com/a/66902167/10049850

backend/graphql/types/userType.ts
```diff
@@ -37,6 +37,7 @@ const userType = gql`
     userByEmail(email: String!): UserDTO!
     users: [UserDTO!]!
     usersCSV: String!
+    userCountByTown(startDate: String, endDate: String): String!
```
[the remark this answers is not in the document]
e-wai: Yes

backend/services/implementations/userService.ts
```diff
@@ -16,6 +17,41 @@ const getMongoUserByAuthId = async (authId: string): Promise<User> => {
...
+const getDateTime = (date: string): string =>
+  `${Math.floor(new Date(date).getTime() / 1000).toString(16)}0000000000000000`;
+
+interface MatchObject {
+  $match: {
+    _id: {
+      [key: string]: Types.ObjectId;
+    };
+  };
+}
```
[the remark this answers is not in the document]
e-wai: ehh I'm not sure about best preference, I kept it here because it's private to this file (they could be anonymous to the same effect, this makes it a tad more readable)

backend/services/implementations/userService.ts
e-wai: re: tests that makes sense, I've been doing that via manual test cases so far (in PR description)
re: comment, is that different from the intuitive usage? especially since we don't actually specify what a "beginning" or "end" date would be in the query

backend/services/implementations/userService.ts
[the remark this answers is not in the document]
e-wai: updated name atm is `LimitByDateQuery`

backend/services/implementations/userService.ts
draymond63: Not sure I fully understand what you're saying, but I think what you have now is good. We can discuss it in today's meeting

backend/services/implementations/userService.ts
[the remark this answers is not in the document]
draymond63: Yeah I think you're right after sleeping on it. There are other files where types are defined inline with the functions that use them.

draymond63 (approved):
Things look good to me. I'd like someone else to review it before merging, but it can merge if everybody else is happy with it :)

backend/services/implementations/userService.ts
```diff
@@ -16,6 +17,50 @@ const getMongoUserByAuthId = async (authId: string): Promise<User> => {
...
+
+interface LimitByDateQuery {
+  $match: {
+    _id: {
+      [key: string]: Types.ObjectId;
+    };
+  };
+}
+
+const buildLimitByDateQuery = (
```
Shrinjay, line 40: I think the greater than/less than query should be built in its own function, since that's going to be reused. I think of this query as just a special case of a greater than/less than query, the only additional logic we have to do is turning the dates into ObjectIds with datetime. That means we should factor out all the logic that is not specific to turning the dates into an objectId into functions. 

```
buildGtQuery(obj, val) {
   // Allows us to handle the case where val is falsey directly in the query builder, which is what builder functions should do imo
   return !!val ? {...obj, $gt: val} : obj;
}

buildLtQuery(obj, val) {
   return !!val ? {...obj, $gt: val} : obj;
}
```

That means we can simplify this function into:
```
buildLimitByDataQuery(startDate, endDate) {
   const matchQuery = {$match: {_id: {} } };
   builtGtQuery(matchQuery, Types.ObjectId(getDateTime(startDate)));
   builtLtQuery(matchQuery, Types.ObjectId(getDateTime(endDate)));
   return matchQuery;
}
```

Might've missed something here, and the code isn't really proper, but what I'm getting at is that building queries naturally invokes the builder pattern and we should use the builder pattern to its maximum, and imo this is what that looks like.

backend/services/implementations/userService.ts
```diff
@@ -16,6 +17,50 @@ const getMongoUserByAuthId = async (authId: string): Promise<User> => {
...
+ *
+ * ObjectID docs:
+ *  https://mongodb.github.io/node-mongodb-native/api-bson-generated/objectid.html
+ * Reference:
+ *  https://kchodorow.com/2011/12/20/querying-for-timestamps-using-objectids/
+ */
+const getDateTime = (date: string): string =>
+  `${Math.floor(new Date(date).getTime() / 1000).toString(16)}0000000000000000`;
+
+interface LimitByDateQuery {
```
Shrinjay, line 32: This interface isn't really limited to Limiting by date, is it? It's just a match query on the id property, that could be used for any kind of query on id no?

backend/services/implementations/userService.ts
Shrinjay, line 40: Second thing on this, I think the logic for building queries should be separated from the user service, because again, it's pretty general and can be reused (i.e wanting to query courses or modules by date published). Only the top-level function (i.e buildLimitByDate) query **could** be specific to a given service. 

I think the best way to handle this is 
1. Create a queryBuilderService that can be reused across different services to create queries on the data model. 
2. To allow us to scope query builder functions only to services that need them, we should export objects of functions that contain different builder functions that each service can use. For example, if we have service A and B, which can only use builder function A and B respectively, but they can both use builder function C, then we'd export the following from the query builder service
``` 
const serviceAQueryBuilders = { queryBulderA, queryBuilderC }
const serviceBQueryBuilders = { queryBuilderB, queryBuilderC }
export default {serviceAQueryBuilders, serviceBQueryBuilders }
```
Then service A would import serviceAQueryBuilders and B would import B builders.

Shrinjay (commented):
These are some pretty high-level architectural comments but I think that we're going to be doing a lot of custom queries in the future for analytics or just different use cases, so making a really good architecture for building queries will be really helpful down the line.

e-wai (commented):
Good points @Shrinjay -- think I'll merge this for now and then make the refactor another PR

PR #52, opened 2022-02-18
Completed mock pages and routing

Shrinjay:
## Implementation Description
Added course editor, admin overview, and course pages based on front-end architecture we defined.

Note that this is a pretty small PR so I only asked for one reviewer. 

Overall structure of the pages and routes. Note that the components are just placeholders. 

## Testing Procedure
Breaking the routing or the url params.

## Things to Note

## Checklist


frontend/src/constants/Routes.ts
```diff
@@ -13,3 +13,9 @@ export const CREATE_ENTITY_PAGE = "/entity/create";
 export const UPDATE_ENTITY_PAGE = "/entity/update";
 
 export const HOOKS_PAGE = "/hooks";
+
+export const COURSE_OVERVIEW_DASHBOARD_BASE_ROUTE = "/course"
```
draymond63, line 17: This route would be for the learner's side correct? Not sure if we need this right now, but I'm down to leave it in.

frontend/src/App.tsx
```diff
@@ -79,6 +84,31 @@ const App = (): React.ReactElement => {
...
+              <PrivateRoute
+                exact
+                path={`${Routes.COURSE_OVERVIEW_DASHBOARD_BASE_ROUTE}/:id`}
+                component={CourseOverview}
+              />
+              <PrivateRoute
+                exact
+                path={Routes.ADMIN_COURSE_EDITOR_BASE_ROUTE}
+                component={CourseEditorOverview}
+              />
```
draymond63, line 106: This raises an interesting question, but I don't think we should have pages for the base routes. For example, I don't think `/admin/course` should be valid, but I think `/admin/course/<id>` should be. Same goes for the course base route. What are peoples thoughts on this? I'm simply not sure how we would use them.

frontend/src/App.tsx
```diff
@@ -10,6 +10,11 @@ import Default from "./components/pages/Default";
 import DisplayPage from "./components/pages/DisplayPage";
 import NotFound from "./components/pages/NotFound";
 import UpdatePage from "./components/pages/UpdatePage";
+import AdminDasboard from "./components/pages/admin/AdminDashboard";
+import CourseDashboard from "./components/pages/course/CourseDashboard";
+import CourseEditor from "./components/pages/course/CourseEditor";
+import CourseEditorOverview from "./components/pages/course/CourseEditorOverview";
+import CourseOverview from "./components/pages/course/CourseOverview";
```
draymond63, line 17: Please rename the Course editor components to Module editors. As well, I don't think CourseEditorOverview page - see comment on lines 102-106 for reasoning.

draymond63 (changes requested):
Thanks for doing this! I know things are very busy. Few nomenclature changes, as well as some thoughts about base routes. Let me know what your thoughts are!

draymond63:
Just as a quick overview - these I believe are the page we will be making this term. Sorry for keeping this stuff up in the air, but I think you have it mostly right. The course/module preview pages are a bit of a question mark so lets leave them for now

frontend/src/App.tsx
Shrinjay: These base routes are for the overview pages, for example /admin/course would show a list of courses. They're in the mockup as the "Manage Courses" tab, but I'm not sure if I misinterpreted it.

frontend/src/constants/Routes.ts
Shrinjay: This would be for the list of courses a user is enrolled in, but it seems that page is archived now. It still seems like it would be a requirement so I think we can leave it in?

frontend/src/App.tsx
Shrinjay: While I am doing this right now, I think we'd still need a course editor component since it seems like we do directly edit courses? I'd think we'd need a module edit page in addition to this.

draymond63 (changes requested):
Things look pretty much perfect - we will discuss the AdminDashboard vs ModuleEditorOverview distinction, but that can be in today's meeting. Only thing I'd ask is if you could fix the linter warnings - it just makes things look nicer and less people will have to be warned about it in the future

Shrinjay:
> Things look pretty much perfect - we will discuss the AdminDashboard vs ModuleEditorOverview distinction, but that can be in today's meeting. Only thing I'd ask is if you could fix the linter warnings - it just makes things look nicer and less people will have to be warned about it in the future

Fixed them, sorry about that, can't believe I missed linter errors

draymond63 (approved):
Merging!

Issue #55, opened 2022-02-20
Create lesson API client


draymond63:
## Summary
Now that we have backend APIs that can be called, we need to create some helper functions so the frontend can easily call them. 

## Goals
- [x] create a `lessonAPIClient.ts` file
- [ ] create helper functions within this file to simplify the process of interfacing with the backend
  - get: retrieves a lesson, given an ID
  - create: creates a new lesson
  - update: updates an existing lesson, given an ID
  - delete: deletes a lesson, given an ID

## Implementation
This code should only be written in the `frontend/src/APIClients/` folder. The new file you create should mimick the functionality of the `EntityAPIClient` quite closely. Currently the error catching is redunant (it catches the error just to throw it again), but please do this as well so that in the future we can implement a better system for all our API clients.

In the future, we would most likely like to implement a `getAll` function, which would retrieve all lessons from a given module, but the backend API is not built for that yet.

PR #59, opened 2022-02-23
Implement Lesson API Client

y2kwak:
## Implementation Description
Adding a lesson api client! 

## Screenshots
N/A

- ~~Do we need the `bearerToken` like in [EntityAPIClient](https://github.com/uwblueprint/rowan-house/blob/main/frontend/src/APIClients/EntityAPIClient.ts#L35-L42)?~~
- Ensure that request and response types are appropriate.

## Testing Procedure

- Need to implement unit tests on all the methods

## Things to Note
_- Need to implement better error handling_

## Checklist

### DIFF
--- a/frontend/src/APIClients/lessonAPIClient.ts
+++ b/frontend/src/APIClients/lessonAPIClient.ts
@@ -21,7 +21,7 @@ export type LessonResponse = {
 
 // TO DO: error handling
 
-const get = async (id: string): Promise<LessonResponse> => {
+const getLesson = async (id: string): Promise<LessonResponse> => {
   const bearerToken = `Bearer ${getLocalStorageObjProperty(
     AUTHENTICATED_USER_KEY,
     "accessToken",
@@ -36,7 +36,7 @@ const get = async (id: string): Promise<LessonResponse> => {
   }
 };
 
-const create = async (lesson: LessonRequest): Promise<LessonResponse> => {
+const createLesson = async (lesson: LessonRequest): Promise<LessonResponse> => {
   const bearerToken = `Bearer ${getLocalStorageObjProperty(
     AUTHENTICATED_USER_KEY,
     "accessToken",
@@ -51,7 +51,7 @@ const create = async (lesson: LessonRequest): Promise<LessonResponse> => {
   }
 };
 
-const update = async (
+const updateLesson = async (
   id: string,
   lesson: LessonRequest,
 ): Promise<LessonResponse> => {
@@ -84,4 +84,4 @@ const deleteLesson = async (id: string): Promise<string> => {
   }
 };
 
-export default { get, create, update, deleteLesson };
+export default { getLesson, createLesson, updateLesson, deleteLesson };
--- a/frontend/src/APIClients/lessonAPIClient.ts
+++ b/frontend/src/APIClients/lessonAPIClient.ts
@@ -2,6 +2,8 @@ import baseAPIClient from "./BaseAPIClient";
 import AUTHENTICATED_USER_KEY from "../constants/AuthConstants";
 import { getLocalStorageObjProperty } from "../utils/LocalStorageUtils";
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 export type LessonRequest = {
   course: string;
   title: string;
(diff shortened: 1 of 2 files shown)


frontend/src/APIClients/lessonAPIClient.ts
```diff
@@ -0,0 +1,61 @@
+import baseAPIClient from "./BaseAPIClient";
+
+export type LessonRequest = {
+  course: string;
+  title: string;
+  description: string;
+  image: string;
+  content: [Record<string, unknown>];
```
Shrinjay, line 10: Wouldn't `any` be more appropriate than `unknown` here?

frontend/src/APIClients/lessonAPIClient.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+// TO DO: error handling
+
+const get = async (id: string): Promise<LessonResponse> => {
+  try {
```
Shrinjay, line 29: I'm pretty sure we need to include the bearer token here like we do with EntityAPIClient, that token will be needed for any authenticated routes which the lessons route should be?

Shrinjay (changes requested):
Bearer token probably does need to be included, otherwise just 1 minor thing

frontend/src/APIClients/lessonAPIClient.ts
```diff
@@ -0,0 +1,89 @@
+import baseAPIClient from "./BaseAPIClient";
```
e-wai, line 1: Can we rename file to `LessonAPIClient.ts` (caps different)?

frontend/src/APIClients/lessonAPIClient.ts
```diff
@@ -0,0 +1,89 @@
...
+export type LessonRequest = {
+  course: string;
+  title: string;
+  description: string;
+  image: string;
+  content: [Record<string, any>];
+};
+
+export type LessonResponse = {
+  id: string | number;
```
e-wai, line 16: won't it always be a string?

frontend/src/APIClients/lessonAPIClient.ts
```diff
@@ -0,0 +1,89 @@
...
+  course: string;
+  title: string;
+  description: string;
+  image: string;
+  content: [Record<string, any>];
+};
+
+// TO DO: error handling
+
+const getLesson = async (id: string): Promise<LessonResponse> => {
```
e-wai, line 26: Do we want any other getters? Off the top of my head,
- `get` -> returns all lessons
- `getLessonsByCourse` -> returns all lessons belonging to a course

frontend/src/APIClients/lessonAPIClient.ts
```diff
@@ -0,0 +1,89 @@
...
+  const bearerToken = `Bearer ${getLocalStorageObjProperty(
+    AUTHENTICATED_USER_KEY,
+    "accessToken",
+  )}`;
+  try {
+    const { data } = await baseAPIClient.get(`/lesson${id}`, {
```
e-wai, line 32: What's the difference between `/lesson/${id}` and `/lesson${id}`? Starter code seems to follow the former

e-wai (changes requested):
Looks pretty good -- couple of questions, but nothing major

frontend/src/APIClients/lessonAPIClient.ts
y2kwak: Good catch! It should be `/lesson/${id}`

frontend/src/APIClients/lessonAPIClient.ts
y2kwak: Currently, we only have the [getLessonById](https://github.com/uwblueprint/rowan-house/blob/main/backend/services/implementations/lessonService.ts#L15) in the back end. It's outlined in[ the ticket ](https://github.com/uwblueprint/rowan-house/issues/55)that retrieving all lessons of a module should be implemented separately
But if it makes sense, maybe this getter should be renamed to `getLessonById`.. lmk what you think!

frontend/src/APIClients/lessonAPIClient.ts
y2kwak: I noticed that the [EntityResponse](https://github.com/uwblueprint/rowan-house/blob/main/frontend/src/APIClients/EntityAPIClient.ts#L21) type implemented it as string or number. 

But yeah, I would assume it'll always be a string so I'll remove the or number

draymond63 (approved):
Looks good to me! I'll let others review, but if they're happy with it, it can be merged

frontend/src/APIClients/lessonAPIClient.ts
e-wai: Meh it's pretty minor to me. Whatever is consistent/clear with other usages in repo

e-wai (approved):
Seems good to me -- how'd you test this?

Shrinjay (approved):

PR #79, opened 2022-03-04
Limit access to courses based on role and private/published attributes

e-wai:
## Implementation Description
Separates access to courses between (User) and (Staff, Admin):
- users only have access to courses marked `private: false` and `published: true`
- staff/admin have access to all courses

Changes resolver to fulfill with appropriate logic
- Chatting with internal tools about this approach -- https://uwblueprint.slack.com/archives/C021C0R7KM2/p1645989506184349

- Is this the right separation of logic across layers?

## Testing Procedure

Not sure how to test at the moment since no test admin users... should check that it filters as expected for various combinations of `private` and `published` attributes

## Things to Note

## Checklist

### DIFF
--- a/backend/graphql/resolvers/courseResolvers.ts
+++ b/backend/graphql/resolvers/courseResolvers.ts
@@ -16,6 +16,7 @@ import IEmailService from "../../services/interfaces/emailService";
 import IUserService from "../../services/interfaces/userService";
 import { Role } from "../../types";
 import { CourseVisibilityAttributes } from "../../models/course.model";
+import { assertNever } from "../../utilities/errorUtils";
 
 const courseService: ICourseService = new CourseService();
 const userService: IUserService = new UserService();
@@ -30,8 +31,10 @@ const getCourseVisibilityAttributes = (
       return { private: false, published: true };
     case "Staff":
       return { published: true };
-    default:
+    case "Admin":
       return {};
+    default:
+      return assertNever(role);
   }
 };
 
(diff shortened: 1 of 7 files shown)


backend/services/implementations/courseService.ts
```diff
@@ -36,9 +36,16 @@ class CourseService implements ICourseService {
...
   }
 
-  async getCourses(): Promise<CourseResponseDTO[]> {
+  async getCourses(onlyPublicCourses: boolean): Promise<CourseResponseDTO[]> {
     try {
-      const courses: Array<Course> = await MgCourse.find();
+      const queryConditions = onlyPublicCourses
+        ? {
+            private: false,
+            published: true,
```
draymond63, line 44: Published is a property of modules, not courses. I believe we will have to query the documents based on private status, and the remove the modules that are not published (for staff). Because of this, I think this PR should wait for Shrinjay's module implementation

draymond63, line 44: Users can see courses when:
- private = false, published = true

Staff can see courses when:
- private = true/false, published = true

Admin can see courses when:
- private = true/false, published = true/false

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -1,12 +1,39 @@
...
+
+const getCourseVisibilityAttributes = (
+  role: Role,
+): CourseVisibilityAttributes => {
+  switch (role) {
+    case "User":
+      return { private: false, published: true };
+    case "Staff":
+      return { published: true };
+    default:
```
e-wai, line 33: need to find a way to indicate that the switch case is exhaustive, sub default case with "Admin"

draymond63 (approved):
Looks good to me. In the future, we will need move the published attribute to modules, not courses. The logic for this PR will also have to be implemented in the `course` API as well, restricting some courses and removing some modules where appropriate.

e-wai:
Can we hold off on merging until the switch case comment is addressed?

Issue #39, opened 2022-02-01
Content validation


draymond63:
## Summary
Content will exist as objects in a list in a `lesson` document. This content needs to be validated to make sure that we are only storing correct content. The goal is to create some helper functions that check to make sure every content object has the correct attributes. These helper functions will then be applied to the lesson CRUD APIs.

## Goals
- [x] Define all possible content block types
  - Ideally all types could be referenced using a generic `ContentBlock` type
  - See [Data Models](https://www.notion.so/uwblueprintexecs/Data-Models-f6b8178406c1481da654a70e30eabe0e#5fe186b9d82544b49dbab5a37be0f1a7) for more info
- [x] Create backend helper functions that will ensure objects in the `content` list are valid
  - One function given a single `ContentBlock`, returns boolean
  - One function given a list of `ContentBlock`, returns boolean
- [x] Update lesson CRUD functions to use validation
  - Make sure to update DTO types to uses `ContentBlock`'s

## Implementation
To start, implement the following possible content block types:
- Text
- Image
- Video

PR #80, opened 2022-03-04
Add Lesson Content Validation

molly-yu:
## Implementation Description
Implements content block types and validation (#39 ), with text, image and video types.

- the new GraphQL types created
- the validation methods for the different content types
- use of the new Content Block type

## Testing Procedure
- Test the GraphQL queries/mutations with the actual content stored in the lessons (can be Text, Image, or Video)
- ex.  
```
mutation{
createLesson(lesson: {
    course: "ID",
    title:"Lesson 1",
    content: [{
      type: "text"
		content: {
			text: "123"
		}
    }],
  }) {
    id,
    course,
    title,
    description,
    image,
    content{
      type
    },
  }
}
```

   Parameters must match those defined in the Content Data Models (different for each type)

## Things to Note

For the GraphQL types, the types were split up according to whether or not it was an input type. Since these types do not support inheritance, I had to put the different types of content parameters (link, text) for different content types in the same type. Maybe there is a better way to do this?
We have the ContentBlock, which has a type String as well as the Content, which is an object containing the actual values for the content (e.g. link for photo, video; text for Text type).

## Checklist


backend/types.ts
```diff
@@ -1,5 +1,24 @@
 export type Role = "User" | "Admin";
```
e-wai, line 2: Don't think all these should exist here _and_  in `backend/types/contentBlockType.ts`?

backend/types/contentBlockType.ts
```diff
@@ -0,0 +1,18 @@
+type contentType = "text" |"image"|"video";
```
e-wai, line 1:
```suggestion
export type ContentType = "text" | "image" | "video";
```

- does it make more sense to place this inline in line3?
- nitpick: spacing, capitalization

e-wai: Also when do we use interfaces vs types? Is types only when we need to take advantage of unions?
(cc @draymond63 if you have a style opinion)

backend/middlewares/validators/lessonValidators.ts
```diff
@@ -27,9 +28,44 @@ export const lessonRequestDtoValidator = async (
...
+};
+
+export const validateContentList = (contentList: ContentBlock[]): boolean => {
+  contentList.forEach((contentBlock) => {
+    if(!validateContent(contentBlock)){
+      return false;
+    }
+  });
+  return true;
+}
```
e-wai, line 71:
```suggestion
export const validateContentList = (contentList: ContentBlock[]): boolean =>
  contentList.some((contentBlock) => !validateContent(contentBlock));
```

(may want to double check that logic)

backend/middlewares/validators/lessonValidators.ts
```diff
@@ -27,9 +28,44 @@ export const lessonRequestDtoValidator = async (
   if (!validatePrimitive(body.image, "string")) {
     return res.status(400).send(getApiValidationError("image", "string"));
   }
-  // TO DO: Write better validator for content
+  // content validation
```
e-wai, line 31: don't think this comment is super helpful

backend/middlewares/validators/lessonValidators.ts
```diff
@@ -27,9 +28,44 @@ export const lessonRequestDtoValidator = async (
...
+  if(!validateContentList(body.content)){
+    return res.status(400).send("There was an error in the lesson content.")
```
e-wai, line 33: Could we have a better error message specifying that the inputs didn't fit type expectations?

backend/types/contentBlockType.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface ContentBlock {
+    type: contentType;
+    content: {}
+}
+
+export interface TextBlock extends ContentBlock {
+    content: {text: String;}
```
e-wai, line 8: Do these still work with `String` instead of `string`?

backend/middlewares/validators/lessonValidators.ts
```diff
@@ -27,9 +28,44 @@ export const lessonRequestDtoValidator = async (
...
+  }
   if (typeof body.content !== "object") {
     return res.status(400).send("There was an error");
   }
   return next();
 };
+
+
+export const validateContent = (content: ContentBlock): boolean => {
+  if (content.type === undefined || content.type === null) return false;
```
e-wai, line 43: typescript question -- would `if (!content.type) return false;` work? Pretty sure the falsy values (including undefined, null) are all not expected values

e-wai: same sort of question for the validation @ ln48, ln52, ln56 but I guess maybe an empty string at `text.content.text` could be accepted

e-wai (changes requested):
some quick comments 👍

backend/types/contentBlockType.ts
draymond63: Tbh that's the way I've been using types. Default interface, types for union logic. I think that's generally the standard, I don't see too much of a reason to deviate

backend/middlewares/validators/lessonValidators.ts
draymond63: I agree with this. Knowing what part of the content failed will probably help us down the line

backend/graphql/resolvers/lessonResolvers.ts
```diff
@@ -1,10 +1,12 @@
+import { GraphQLScalarType, Kind } from "graphql";
```
draymond63, line 1: Please resolve these linter errors. Some things it fixes automatically (like formatting), but things like unused imports it doesnt.

backend/graphql/types/lessonType.ts
```diff
@@ -1,33 +1,49 @@
...
+
+  type ContentBlock {
+    type: String!
+    content: Content!
+  }
+
+  input ContentBlockInput {
+    type: String!
+    content: ContentInput!
+  }
```
draymond63, line 22: I'm not familiar enough with how GraphQL works to know really what's going on here. Could you explain?

backend/package.json
```diff
@@ -80,6 +80,6 @@
     "graphql-upload": "^11.0.0"
   },
   "engines": {
-    "node": "14.15.5"
+    "node": "*"
```
draymond63, line 83: I'm not sure this is a change we want to make. I know it can sometimes cause difficulty and throw errors when it isn't a big deal, but I don't think we should commit this change. Could you revert this?

backend/middlewares/validators/util.ts
```diff
@@ -39,6 +41,38 @@ export const validateArray = (value: any, type: Type): boolean => {
   );
 };
 
+export const validateContent = (content: ContentBlock): boolean => {
+  if (content.type === undefined || content.type === null) return false;
```
draymond63, line 45: I agree with Emma: `if (!content.type) return false;` is sufficient I think

backend/types.ts
draymond63: Are they? Did Molly fix this? I don't see a `backend/types/contentBlockType.ts` file in this PR

draymond63 (changes requested):
Thanks for doing this! I know you've been really busy.

There's some comments and questions here and there. I'm curious to know how the GraphQL types work :)

backend/package.json
molly-yu: Yep for sure, I think this change accidentally made it into the commit. I initially changed this because I was running into some node compatibility issues earlier .

backend/types.ts
molly-yu: Hmm that was probably an older version ^

backend/graphql/types/lessonType.ts
molly-yu: I have a small blurb on this in the "Things to Note" section of the PR description. Lmk if you have any specific questions!

backend/types.ts
e-wai: Yeah I probably commented on the wrong line

draymond63 (approved):
Things look good to me!

As long as things have been tested, I think it's good to merge

e-wai (approved):
@draymond63  said tested works, that's the only thing I was holding out on

Issue #69, opened 2022-02-24
Login Flow: Create forgot-password dynamic routing


draymond63:
## See User Flow at: #36 
This task focuses on backend support for the forgot-password pages. The user must verify their email first before being able to change their password. This means that we must send them a special link that takes them to the password-changing page. 

## Goals
- [ ] Dynamically generate link to send to user in email that sends them to the password-changing page

## Implementation
The google-auth API must have a way to create a link that will reset the users password. We would like to create this link and send it in some way to the user via email. When they click the link, it should take them to the password-changing page, which should already be created.

Issue #68, opened 2022-02-24
Login Flow: Create forgot-password flow


draymond63:
## See User Flow at: #36 
This task focuses on the forgot-password pages. The main parts to this page are the form, and the image on the right. See the [User Authentication Figma](https://www.figma.com/file/9KqGifATPcKRQytJBqAKeJ/User-Authentication?node-id=2%3A2) for the design. 

## Pages to Build
![image.png](https://images.zenhubusercontent.com/6053e4ca8db077e41d15aa4a/c530f32f-d788-4f95-ac5f-6307ed146606)
![image.png](https://images.zenhubusercontent.com/6053e4ca8db077e41d15aa4a/77e50b8c-792b-45af-8064-0a8cc46a8cb3)

## Goals
- [x] Build email-entering page
- [ ] Build password-changing page

## Implementation
The layout to this page will be nearly identical to the login/signup page, so copy that if they exist.

These are two separate pages. The email-entering page should be the page users are sent to when the click "forgot password". The password-changing page can exist at a new url temporarily, but it's important to know that *users will not be sent to it directly*. They will have to click a link in the email to see that page. The backend/email functionality will be made in a separate ticket, so for now just create the page and "fake" sending the form data somewhere.

The main components to use in the design will be `Text`, `Image`, `Form`, `Input`. To position the image to the right of the form and have the form centered, you'll need to use layouts component. Look into `Grid` or `Flex` possibly to achieve this. 

### As always...
This code should only be written in the frontend. Store any images in an assets directory. Focus on Desktop, but things shouldn't complete break when screen dimensions change. Play around with different browsers and screen dimensions to make sure things always look good. If you are unsure, talk with the designer or Daniel :)

Images can be copied from Figma by clicking on the layer then clicking the export tab. Images should be exported to PNG, whereas vector graphics (e.g. made in Figma) should be exported to SVG.

Search up Chakra components at https://chakra-ui.com/.

PR #81, opened 2022-03-04
Frontend of Login Flow

binalpreetkalra:
## Implementation Description
Create the frontend pages for the login page with the form and image on the right.

## Screenshots
image
image

- Ensuring this aligns with the Figma file -> https://www.figma.com/file/9KqGifATPcKRQytJBqAKeJ/User-Authentication?node-id=316%3A3

## Testing Procedure
- Able to input an email address and password
- Button text changes once email is entered
- Basic formatting of the page is reflective of the Figma on different screens

## Checklist


frontend/src/components/auth/Login.tsx
```diff
@@ -25,17 +40,24 @@ const Login = (): React.ReactElement => {
...
 
   const onLogInClick = async () => {
-    const user: AuthenticatedUser = await authAPIClient.login(
-      email,
-      password,
-      login,
-    );
-    setAuthenticatedUser(user);
+    if (isFirstPage === false) {
+      setIsFirstPage(true);
```
DinuWije, line 50: whats the point of using the setter here?

frontend/src/components/auth/Login.tsx
```diff
@@ -25,17 +40,24 @@ const Login = (): React.ReactElement => {
...
+    } else if (isFirstPage === true) {
+      const user: AuthenticatedUser = await authAPIClient.login(
+        email,
+        password,
+        login,
+      );
+      setAuthenticatedUser(user);
+    } else {
```
DinuWije, line 58: do we need an `else if` statement for this condition?

frontend/src/components/auth/Login.tsx
```diff
@@ -47,45 +69,72 @@ const Login = (): React.ReactElement => {
...
+          flex='1'
         >
-          Sign Up
-        </button>
-      </div>
-    </div>
+          <VStack
+          >
+          <Image htmlHeight="5" src={logo} />
+          <Heading>Sign in to access courses</Heading>
```
DinuWije, line 80: where is the styling done for these components?

frontend/src/components/auth/Login.tsx
```diff
@@ -25,17 +40,24 @@ const Login = (): React.ReactElement => {
...
+    if (isSecondPage === false) {
+      setIsSecondPage(true);
+    } else if (isSecondPage === true) {
+      const user: AuthenticatedUser = await authAPIClient.login(
+        email,
+        password,
+        login,
+      );
+      setAuthenticatedUser(user);
+    } else {
```
draymond63, line 58: I think since we have two separate button components, it'd be great two separate the logic into two functions, one for each button. That we we can avoid this conditional statements and just run what we want based on the buttons

frontend/src/components/auth/Login.tsx
```diff
@@ -47,45 +69,85 @@ const Login = (): React.ReactElement => {
...
-          Sign Up
-        </button>
-      </div>
-    </div>
+    <Flex>
+      <Center flex="1">
+        <VStack>
+          <Image height="13vh" marginBottom="0.5rem" src={logo} />
+          <Text variant="display-sm-sb">Sign in to access courses</Text>
+          <form>
```
draymond63, line 77: The form tag is causing weird default behaviour - when I press enter while in the input it attempts to submit the form. This normally would be good, but we don't want it to do that. Either preventing the default behaviour, or perhaps just getting rid of the form tag would be a good idea.

frontend/src/components/auth/Login.tsx
```diff
@@ -47,45 +69,85 @@ const Login = (): React.ReactElement => {
...
+            {isSecondPage ? (
+              <>
+                <FormControl>
+                  <Text
```
draymond63, line 81: Please resolve the linter issues :)

frontend/src/components/auth/Login.tsx
```diff
@@ -47,45 +69,85 @@ const Login = (): React.ReactElement => {
...
-          onClick={onSignUpClick}
-        >
-          Sign Up
-        </button>
-      </div>
-    </div>
+    <Flex>
+      <Center flex="1">
+        <VStack>
+          <Image height="13vh" marginBottom="0.5rem" src={logo} />
```
e-wai, line 75: nitpick: maybe keep consistent with the units?

frontend/src/components/auth/Login.tsx
```diff
@@ -1,6 +1,21 @@
...
+  FormLabel,
+  Link,
+  Box,
+  Button,
+  VStack,
+  Input,
+  Center,
+} from "@chakra-ui/react";
+import logo from "../logo.png";
+import background from "../signuppage.png";
```
e-wai, line 18: super nit: capitalize the name, and include "image" or "png" in the name? eg. `LogoBackgroundPng`

frontend/src/components/auth/Login.tsx
```diff
@@ -47,45 +69,85 @@ const Login = (): React.ReactElement => {
...
+                <FormControl>
+                  <Text 
+                    variant="caption-md"
+                    marginTop="1.5rem"
+                    > Password</Text>
+                  <Input
+                    type="password"
+                    placeholder="●●●●●"
+                    marginBottom="1rem"
+                    onChange={(event) => setPassword(event.target.value)}
```
draymond63, line 89: For some reason, when I go through the form login, it auto-populates the password which a bunch of characters. Not sure why this is, but it'd be good to get rid of. It's not the placeholder, but something is there that I have to delete manually. Are other people experiencing this?
image

frontend/src/styles/colors.ts
```diff
@@ -0,0 +1,31 @@
+const colors = {
```
draymond63, line 1: All of of these colors are in `frontend/src/theme/foundations/colors.ts`. Could we delete this to avoid duplication? How are you using it?

frontend/src/styles/theme.ts
```diff
@@ -0,0 +1,19 @@
...
+    // body: "Inter, sans-serif",
+  },
+  // textStyles,
+  colors,
+  config: {
+    initialColorMode: "light",
+  },
+});
+
+export default customTheme;
```
draymond63, line 19: I think these files are relics from when you implemented the chakra theme - could we delete this folder?

draymond63 (changes requested):
Thanks binal! Just a few comments, but overall great work! I hope you've learned a lot doing this, it's a big accomplishment!

frontend/src/components/auth/Login.tsx
```diff
@@ -25,17 +40,24 @@ const Login = (): React.ReactElement => {
   const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
+  const [isSecondPage, setIsSecondPage] = useState<boolean>(false);
```
e-wai, line 43: I think we need (1) more descriptive names for this and (2) more than just two states, so maybe a type would work better? Something like

```ts
const [loginState, setLoginState] = useState<LoginState>(LoginState.EnterEmail);

interface LoginState {
  EnterEmail,
  EnterPassword,
  ForgotPassword,
};
```

and then getters to return the appropriate details? As in

```ts
const getBottomPrompt = (loginState: LoginState): React.ReactElement => {
  switch(loginState) {
    case LoginState.EnterEmail:
      return(<Text>Don't have an account? <Link>Sign up</Link></Text>);
    case:
      ...

  }
}
```

(I just typed this out of VScode so not sure it compiles, but hopefully the idea is somewhat clear)

frontend/src/components/auth/Login.tsx
```diff
@@ -47,45 +69,85 @@ const Login = (): React.ReactElement => {
...
+                >
+                  Login
+                </Button>
+                <Link
+                  color="purple"
+                  onClick={onSignUpClick}
+                  href="https://www.figma.com/file/9KqGifATPcKRQytJBqAKeJ/User-Authentication?node-id=316%3A2"
+                >
+                  Forgot your password
+                </Link>
```
e-wai, line 106: What is the `href` field for if it's already handled by `onSignUpClick`?

frontend/src/components/auth/Login.tsx
e-wai: (Similar logic would also replace all the `isSecondPage ?` conditionals in the return func.) 

nit: It's also possible that we extract all each "return" clause into its own component... while that means more repeated code, it might also be more readable than having this much branching. I don't think there's too much downside either way, but maybe other people have suggestions? 

cc: @y2kwak @draymond63

frontend/src/styles/colors.ts
```diff
@@ -0,0 +1,31 @@
...
+    100: "2D3748",
+  },
+  grey: {
+    100: "6C6C84",
+  },
+  critical: {
+    100: "E53E3E",
+  },
+};
+export default colors;
```
e-wai, line 31: Think this already exists

frontend/src/styles/theme.ts
e-wai, line 19: Do we have plans for a dark mode?

e-wai (changes requested):
Hey, this looks great so far good job 🎉 

I have a couple questions about structure and I think this needs a rebase to pick up on some of the new style, theme settings. Let me know if anything is unclear!
- I think the `logo.png` was added in the repo already as `RHSlogo.png` but might need a rebase to find it

e-wai:
Visual nitpicks:

Screen Shot 2022-03-24 at 7 03 48 PM

- [ ] Looks like logo is too big
- [ ] Font too big on "email address" input label
- [ ] Underline on links
- [ ] "Forgot your password" should be centered
- [ ] No "back" button right now on the enter password screen
- [ ] Maybe run this by design to see what they think?
  - [ ] What do they want responsiveness-wise?

frontend/src/components/auth/Login.tsx
e-wai: This one's a bit weird, when I enter 4 characters in the email, there are 4 dots; same for other #s. Is it maybe picking up on the same input field?

frontend/src/components/auth/Login.tsx
```diff
@@ -21,71 +36,215 @@ const LOGIN = gql`
...
+  const onLogInClick = async () => {
+    switch (loginState) {
+      case LoginState.EnterEmail:
+        setLoginState(LoginState.EnterPassword);
+        break;
+      case LoginState.EnterPassword:
+        getLoginUser();
+        break;
+      case LoginState.ForgetPassword:
+        // implement send forget password link
```
e-wai, line 73: Can we create a task for this?

e-wai (approved):
This looks great, wow!! 🔥 

I had one more small comment + did we check this with design? Otherwise, happy to merge but maybe others have outstanding questions

frontend/src/components/auth/Login.tsx
draymond63: There is a ticket I believe, #68 & #69

draymond63 (approved):
I'm happy with it! We can merge whenever you are happy with it too :)

frontend/src/components/auth/Login.tsx
```diff
@@ -21,71 +36,218 @@ const LOGIN = gql`
...
+        return (
+          <Box>
+            <Text variant="display-sm-sb">Sign in to access courses</Text>
+            <FormControl>
+            <Box 
+              display="flex"
+              marginTop="3vh"
+              >
+              <Button
+                onClick={() => onBackClick}
```
draymond63, line 150:
```suggestion
                onClick={() => onBackClick(loginState)}
```

PR #87, opened 2022-03-06
Fix Lesson API Client - Update call parameter type

y2kwak:
## Implementation Description
Small fix to the updateLesson method in the Lesson API Client. This is to allow the lesson update request to be a partial type of the LessonRequest type.

## Screenshots
N/A

- 

## Testing Procedure

## Things to Note

## Checklist


e-wai (approved):
Code looks fine 👍 

(Commit history looks weird, not that it matters with the squash + merge. Maybe squash commits + force push just to clean that up?)

PR #90, opened 2022-03-07
Add custom text variants/styles, utils command for linting

e-wai:
## Implementation Description
Initial few changes for course block ticket
- Only separating this into its own ticket to make sure the text stylings are available for all future frontend work (and so no duplication on this work)

- Please compare to the Figma -> https://www.figma.com/file/zh6zk2LJo0RodIHBVycylC/RHS-Design-System?node-id=217%3A9061
  - All styles should align exactly with values in `text.ts`
- What the heck is the `em` property supposed to be 

## Testing Procedure

Utils script should be manually testable
-> on Mac, command is `bash utils.sh felint` // `bash utils.sh belint`
-> may not work on other platforms

## Things to Note
- Things you'd want to know when coming back to the code after a few months -> find a fix for other platforms?

## Checklist

### DIFF
--- a/frontend/src/theme/components/index.ts
+++ b/frontend/src/theme/components/index.ts
@@ -1,4 +1,5 @@
 import Button from "./button";
 import Input from "./input";
+import Text from "./text";
 
-export { Button, Input };
+export { Button, Input, Text };
--- a/frontend/src/theme/components/text.ts
+++ b/frontend/src/theme/components/text.ts
@@ -1,0 +1,68 @@
+const Text = {
+  baseStyle: {},
+  sizes: {},
+  variants: {
+    "display-lg": {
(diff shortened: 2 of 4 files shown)


e-wai:
Actually maybe the button styles are unnecessary here and should be manually added into `button.ts`?

draymond63 (commented):
Is the Chakra `Text` component the way we should be handling text styles? Perhaps can include these as general "fonts" and then we can apply them to whatever container we wish (e.g. `Text` or `Header` or `div`)

If we decide to go with Text components, I'm happy with this.

draymond63 (approved):
After reading through Chakra documentation, this is probably the best way to go. I'm happy!

PR #92, opened 2022-03-11
Implement Modules

Shrinjay:
## Implementation Description
This PR involves adding modules to the database as a layer between courses and lessons, modules are added as a property of course objects. 

- Ensure that my validation logic is valid, I built a custom object to validate objects in the future which might come in handy
- Ensure that the serialization stuff won't break
- Ensure that the modules data model and all is implemented correctly

## Testing Procedure
- Try passing any invalid inputs 

## Checklist


backend/package.json
```diff
@@ -29,6 +29,7 @@
     "firebase-admin": "^9.5.0",
     "graphql": "^15.5.0",
     "graphql-middleware": "^6.0.6",
+    "graphql-type-json": "^0.3.2",
```
draymond63, line 32: This is no longer true right? Could you remove this from the commit?

backend/services/interfaces/ICourseService.ts
```diff
@@ -15,9 +25,59 @@ export interface UpdateCourseRequestDTO {
...
+  title: string;
+  description: string;
+  image: string;
+  previewImage: string;
+  private: boolean;
+  published: boolean;
+  modules: ModuleDTO[];
+}
+
+export interface SerializedCourseResponseDTO {
```
draymond63, line 62: I wonder if we could extend/partial the regular DTOs, and simply replace the modules attribute with the correct type. Is that something we could do? It would definitely cut down on the duplication of code.

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -44,14 +105,16 @@ const courseResolvers = {
...
       _parent: undefined,
       { id }: { id: string },
-    ): Promise<CourseResponseDTO> => {
-      return courseService.getCourse(id);
+    ): Promise<SerializedCourseResponseDTO> => {
+      return courseService
+        .getCourse(id)
+        .then((course) => serializeCourseResponse(course));
     },
     courses: async (
```
draymond63, line 113: Some small linter warnings, could you either resolve them or disable them?

draymond63 (changes requested):
Good work!

The logic is all there and I'm happy with the implementation. Just some nitpicks here and there :)

backend/middlewares/validators/courseValidators.ts
```diff
@@ -1,11 +1,22 @@
...
 import { getErrorMessage } from "../../utilities/errorUtils";
 
+const VALID_MODULES = {
+  id: {
+    title: "string",
+    description: "string",
+    image: "string",
+    preview_image: "string",
+    published: "boolean",
+    lessons: "string",
```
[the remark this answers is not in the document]
Shrinjay: It is, the validator should be able to detect that it's an array and cast this over an array but I realize this is a bit messy so I'll find a better way to mark something as an array.

backend/services/interfaces/ICourseService.ts
Shrinjay: We definitely could, I don't really know why I did this tbh

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -16,14 +21,70 @@ import UserService from "../../services/implementations/userService";
...
 const emailService: IEmailService = new EmailService(nodemailerConfig);
 const authService: IAuthService = new AuthService(userService, emailService);
 
+function serializeToModuleDTO(modules: { [id: string]: Module }): ModuleDTO[] {
+  return Object.entries(modules).map(([key, val]) => ({ id: key, ...val }));
+}
+
+function deserializeModuleDTO(
+  moduleDTOs: ModuleDTO[],
+): { [id: string]: Module } {
```
e-wai, line 37: Think we should make `{ [id: string]: Module }` its own type, it's used a decent amount and is something that should be very clear

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -16,14 +21,70 @@ import UserService from "../../services/implementations/userService";
...
+    const moduleObj = {
+      title: moduleDTO.title,
+      description: moduleDTO.description,
+      image: moduleDTO.image,
+      previewImage: moduleDTO.previewImage,
+      published: moduleDTO.published,
+      lessons: moduleDTO.lessons,
+    };
+
+    // If no UUID is defined, create one
```
e-wai, line 50: nitpick: ehh I think this is fairly self-explanatory, don't know if the comment is really that helpful

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -16,14 +21,70 @@ import UserService from "../../services/implementations/userService";
...
+    modules[moduleDTO?.id ?? uuidv4()] = moduleObj;
```
e-wai, line 51: When do we expect the `moduleDTO` to have an ID vs. not have an ID? If it should always have an ID, we should expect to always have an ID and handle this more forcefully rather than absorbing the error (whether this is showing an error, or making a mutation in the db to fix)

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -16,14 +20,70 @@ import UserService from "../../services/implementations/userService";
...
+  if (course?.modules) {
+    return {
+      ...course,
+      modules: serializeToModuleDTO(course.modules),
+    };
+  }
+  return {
+    ...course,
+    modules: [],
+  };
```
e-wai, line 69: - Doesn't `course` always exist? Why the `?` operator

Would this work?
```suggestion
  return {
    ...course,
    modules: course.modules ? serializeToModuleDTO(course.modules) : [],
  };
```

e-wai: Also, shouldn't `modules` always exist? It might just be empty (I don't remember how it's implemented right now, but initially the `lessons` parameter was required on create course)

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -16,14 +20,70 @@ import UserService from "../../services/implementations/userService";
...
+    return {
+      ...course,
+      modules: deserializeModuleDTO(course.modules),
+    };
+  }
+  return {
+    ...course,
+    modules: {},
+  };
+}
```
e-wai, line 85: same comments as on `serialize` (less verbose version, unexpected usages of `?`)

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -62,22 +124,32 @@ const courseResolvers = {
...
-      return courseService.updateCourse(id, course);
+      { id, course }: { id: string; course: SerializedUpdateCourseRequestDTO },
+    ): Promise<SerializedCourseResponseDTO | null> => {
+      return courseService
+        .updateCourse(id, deserializeCourseRequest(course))
+        .then((serializedCourse) => {
+          if (serializedCourse)
+            return serializeCourseResponse(serializedCourse);
+          return null;
+        });
```
e-wai, line 152: In terms of naming, are we expecting a `serializedCourse` response? I don't think so, because we would serialize it when it's returned -- think the param name on ln#148 doesn't make sense then

nitpick syntax again, does this work?

```suggestion
        .then((serializedCourse) => serializedCourse ? seralizeCourseResponse(serailizedCourse) : null;
```

backend/graphql/types/courseType.ts
```diff
@@ -1,13 +1,33 @@
...
 
 const courseType = gql`
+  type ModuleDTO {
+    id: ID!
+    title: String!
+    description: String
+    image: String
+    previewImage: String
+    published: Boolean!
+    lessons: [String]
```
e-wai, line 11: If this is a return type, shouldn't they all be non-optional fields? Doesn't mean the user has to ask for them but they have to all exist

nitpick: for consistency, `ModuleResponseDTO`

backend/graphql/types/courseType.ts
```diff
@@ -1,13 +1,33 @@
...
+  }
+
+  input ModuleInputDTO {
+    id: ID
+    title: String!
+    description: String
+    image: String
+    previewImage: String
+    published: Boolean!
+    lessons: [String]
```
e-wai, line 21: Think lessons should always exist, but it can be empty. Also, should always be of type string:
```suggestion
    lessons: [String!]!
```

backend/graphql/types/courseType.ts
```diff
@@ -17,7 +37,7 @@ const courseType = gql`
     description: String
     image: String
     previewImage: String
-    lessons: [ID!]!
+    modules: [ModuleInputDTO]
```
e-wai, line 40: Also think these should be `modules: [ModuleDTO!]!`

backend/graphql/types/courseType.ts
```diff
@@ -1,13 +1,33 @@
...
     id: ID!
     title: String!
     description: String
     image: String
     previewImage: String
-    lessons: [ID!]!
+    modules: [ModuleDTO]
     private: Boolean!
     published: Boolean!
   }
```
e-wai, line 33: same comment as on `ModuleDTO`, if it's a response I think they should always be required

backend/graphql/types/courseType.ts
e-wai: also, nitpick: `ModuleInputDTO` -> `ModuleRequestDTO`

e-wai (changes requested):
I'll split the review into two, this is the first half (got somewhat nitpicky, sorry)

backend/middlewares/validators/util.ts
```diff
@@ -39,6 +39,26 @@ export const validateArray = (value: any, type: Type): boolean => {
   );
 };
 
+export const validateObject = (
+  object: { [field: string]: any },
+  types: { [field: string]: any },
+): boolean => {
+  return (
+    object !== undefined &&
+    object !== null &&
```
e-wai, line 48: If I'm remembering my falsy values right, this can all become `return (object) && ...`

backend/middlewares/validators/util.ts
```diff
@@ -39,6 +39,26 @@ export const validateArray = (value: any, type: Type): boolean => {
...
+    typeof object === "object" &&
+    Object.entries(object).every(([key, value]) => {
+      if (key.slice(-2) === "[]" && Array.isArray(value)) {
+        return validateArray(value, types[key.slice(0, -2)]);
+      }
```
e-wai, line 53: Bit unclear on this piece of logic, what's happening here?

backend/graphql/resolvers/courseResolvers.ts
Shrinjay: Yeahh I should've done that, just got lazy, I'll make it its own type

backend/graphql/resolvers/courseResolvers.ts
Shrinjay: It kind of breaks the pattern of generating IDs via MongoDB, but this needs to happen since the modules don't have their own objectID like courses do as they're not in their own collection, that's why I made this a comment.

backend/graphql/resolvers/courseResolvers.ts
Shrinjay: Module DTO might not have an ID when we have a request to create a new course with a new module, in which case we create the ID here, but I agree that doing this more forcefully elsewhere might be cleaner. The other option is generating an ID in the client but that's even more anti-pattern.

backend/graphql/resolvers/courseResolvers.ts
Shrinjay: I wasn't sure, iirc lessons was optional in the graphql type previously so I made modules optional as well, and I think it makes sense to have a newly created course with no modules. Although now that I think about it, it makes more sense to remove this handling and have the frontend always send `[]` to indicate no modules.

backend/graphql/resolvers/courseResolvers.ts
Shrinjay: Yeah good catch, I just messed up the naming

backend/graphql/types/courseType.ts
Shrinjay: I made them optional because the pattern seemed to be that most fields were left optional, didn't really think through it, but yes making them non-optional makes more sense

backend/middlewares/validators/util.ts
Shrinjay: Basically, `string[]` represents an array, so if the last 2 chars of a string are '[]', then we should use validate array. When passing into validateArray, we only want the type of each item, so we remove the `[]` characters.

backend/graphql/resolvers/courseResolvers.ts
e-wai: OK fair enough

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -16,14 +20,70 @@ import UserService from "../../services/implementations/userService";
...
-import { CourseVisibilityAttributes } from "../../models/course.model";
+import { CourseVisibilityAttributes, Module } from "../../models/course.model";
 import { assertNever } from "../../utilities/errorUtils";
 
 const courseService: ICourseService = new CourseService();
 const userService: IUserService = new UserService();
 const emailService: IEmailService = new EmailService(nodemailerConfig);
 const authService: IAuthService = new AuthService(userService, emailService);
 
+function serializeToModuleDTO(modules: { [id: string]: Module }): ModuleDTO[] {
```
y2kwak, line 31: nit: probably doesn't affect logic; but consider using const arrow functions to keep it consistent w the rest of the file

frontend/src/APIClients/CourseAPIClient.ts
```diff
@@ -79,7 +137,7 @@ const updateCourse = async (
     "accessToken",
   )}`;
   try {
-    const { data } = await baseAPIClient.put(`/course/${id}`, course, {
+    const { data } = await baseAPIClient.put(`/course/${id}`, serializeUpdateCourseRequest(course), {
       headers: { Authorization: bearerToken },
     });
     return data;
```
y2kwak, line 143: I might not be following correctly, but do we need to deserialize the course response here?

y2kwak (commented):
this seemed like a complicated problem ngl, but your solution makes sense - good job :))

backend/middlewares/validators/util.ts
e-wai: Makes sense, maybe worth a comment

frontend/src/APIClients/CourseAPIClient.ts
Shrinjay: Yes we do, good catch!

backend/graphql/types/courseType.ts
```diff
@@ -24,10 +24,10 @@ const courseType = gql`
...
     id: ID!
     title: String!
-    description: String
-    image: String
-    previewImage: String
-    modules: [ModuleDTO]
+    description: String!
+    image: String!
+    previewImage: String!
+    modules: [ModuleResponseDTO]!
```
e-wai, line 30: Why not `[ModuleResponseDTO!]!` -- ie. why no strict type check on `ModuleResponseDTO`, do we not expect that the response will always be of type `ModuleResponseDTO`? If it's because of missing ID, I'd rather we make id optional on the other DTO object

backend/graphql/types/courseType.ts
```diff
@@ -37,17 +37,17 @@ const courseType = gql`
     description: String
     image: String
     previewImage: String
-    modules: [ModuleInputDTO]
-    private: Boolean
-    published: Boolean
+    modules: [ModuleRequestDTO]!
```
e-wai, line 40: Same Q as above

backend/graphql/types/courseType.ts
```diff
@@ -37,17 +37,17 @@ const courseType = gql`
...
+    published: Boolean!
   }
 
   input UpdateCourseRequestDTO {
     title: String
     description: String
     image: String
     previewImage: String
-    modules: [ModuleInputDTO]
+    modules: [ModuleRequestDTO]
```
e-wai, line 50: Same ^

frontend/src/APIClients/CourseAPIClient.ts
```diff
@@ -13,12 +13,14 @@ type Module = {
 
 type SerializedModule = Module & {id: string}
 
+type Modules = {[id: string]: Module}
```
e-wai: nitpick: would it make sense to call this `ModulesById`? It makes the usage much more obvious

e-wai (approved):
Looks great 🥳 , some tiny comments left but nothing blocking

- I didn't get around to testing this manually (but I can if no one else has checked it out)
- Could you add the link to the data models updated page in the PR description and add some context about why the serialization is necessary, other non-obvious details about when we expect id, etc.?
- Maybe just check with @y2kwak to see if she's waiting for this to merge before/after the GraphQL stuff? (I think after, but maybe just clear that up to avoid merge conflicts)

draymond63 (approved):
Things look good to me, although Emma did have some good comments about enforcing the types within arrays. There's also a few linter things, but I'm good with this PR getting merged to prevent other things from getting blocked!

frontend/src/APIClients/CourseAPIClient.ts
Shrinjay: That works too

backend/graphql/types/courseType.ts
Shrinjay: Honestly I just didn't know that you needed to do [ModuleResponseDTO!]!

Shrinjay:
Just went ahead and merged this after it built, I fixed what Emma talked about

PR #94, opened 2022-03-12
Fix: button font styling

e-wai:
## Implementation Description
- Change the font styling in the `button.ts` theme customization

## Screenshots
Screen Shot 2022-03-11 at 9 15 32 PM

- Check that this aligns with the Figma file -> https://www.figma.com/file/zh6zk2LJo0RodIHBVycylC/RHS-Design-System?node-id=210%3A13900
  - Was padding right? It's not consistent between the Figma and here, but pretty consistent within the file
- Check that this is the proper way of styling components

## Testing Procedure

Can add into any file and check that the appearances look write
```ts
import { Button } from "@chakra-ui/react";

<Button variant="default">Default</Button>
<Button variant="md">md</Button>
<Button variant="sm">sm</Button>
<Button variant="xs">xs</Button>
<Button variant="disabled">disabled</Button>
<Button variant="outline-lg">outline lg</Button>
<Button variant="outline-md">outline md</Button>
<Button variant="outline-sm">outline sm</Button>
<Button variant="outline-xs">outline xs</Button>
<Button variant="outline-disabled">outline disabled</Button>
```

## Things to Note

None

## Checklist


draymond63 (approved):
Tested, works. Feel free to merge :)

PR #104, opened 2022-03-18
Use Apollo and gql for BaseAPI, Auth, Entity and Courses

y2kwak:
## Implementation Description
Numerous API client fixes:
- Use `ApolloClient` in BaseAPIClient and leverage `ApolloProvider`
- Add AuthMutations file for Auth related gql strings
- Add EntitiesMutations and EntitiesQueries
- Remove LessonAPIClient and CourseAPIClient

## Screenshots
N/A

## Testing Procedure
- I tried all the Entities (create, update) and Auth (login, logout) functionality and it's working as expected

## Things to Note

## Checklist


frontend/src/APIClients/mutations/EntityMutations.ts
```diff
@@ -0,0 +1,33 @@
+import { gql } from "@apollo/client";
```
e-wai, line 1: Ehh I think we're reaching the point that we can start removing the entity stuff... or at the very least not touch it because it's adding unnecessary dev work

frontend/src/APIClients/queries/CourseQueries.ts
```diff
@@ -0,0 +1,29 @@
+import { gql } from "@apollo/client";
+
+export const COURSES = gql`
+  query Courses {
+    courses {
+      title
```
e-wai, line 6: Think we can return `id` here?

frontend/src/APIClients/types/CourseClientTypes.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const serializeUpdateCourseRequest = (
+  courseRequest: Partial<CourseRequest>,
+): SerializedUpdateCourseRequest => {
+  return {
+    modules:
+      courseRequest.modules === undefined
+        ? undefined
+        : serializeModules(courseRequest.modules),
```
e-wai, line 88: How does this handle re: typing? As in, code pathways seem to have a "return undefined" path but the method signature does not

frontend/src/APIClients/BaseAPIClient.ts
```diff
@@ -1,49 +1,113 @@
...
+  if (graphQLErrors)
+    graphQLErrors.forEach(({ message, locations, path }) =>
+      // eslint-disable-next-line no-console
+      console.log(
+        `[GraphQL error]: Message: ${message}, Location: ${locations}, Path: ${path}`,
+      ),
+    );
+  // eslint-disable-next-line no-console
+  if (networkError) console.log(`[Network error]: ${networkError}`);
+});
```
e-wai, line 92: Is this a temporary or long-term error handling approach?

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -131,14 +131,23 @@ type RefreshFunction = (
...
-    setLocalStorageObjProperty(AUTHENTICATED_USER_KEY, "accessToken", token);
+  try {
+    const result = await refreshFunction({ variables: {} });
+    const token = result.data?.refresh;
+    if (token) {
+      setLocalStorageObjProperty(AUTHENTICATED_USER_KEY, "accessToken", token);
+    } else {
+      return false;
+    }
+    return true;
```
e-wai, line 142: nit: simplify flow

```suggestion
    if (!token) {
      return false;
    } 
    setLocalStorageObjProperty(AUTHENTICATED_USER_KEY, "accessToken", token);
    return true;
```

e-wai (commented):
It's soo clean thank you!! 😍 just some minor items here + there 

(Also, I don't think that the entity changes are not worth making)

draymond63 (changes requested):
Main thing is to remove the Apollo provider from index.tsx, as you now have it in the App.tsx file. Thanks for getting this done so quickly! It'd be great if we could have a call to discuss before merging

frontend/src/APIClients/queries/CourseQueries.ts
```diff
@@ -0,0 +1,29 @@
...
+      description
+      image
+      previewImage
+      lessons
```
draymond63, line 10: Courses now contain modules, not lessons (same for the GET_COURSE graphql string)

frontend/src/APIClients/types/CourseClientTypes.ts
y2kwak: Hm,, this is actually extracted from the [implement Modules PR](https://github.com/uwblueprint/rowan-house/pull/92/files#diff-7c71de3be3e3b0b16555de131be1a187b67a7b94367a1da8b9721cb98d1cb68aR84-R93). I'm just reorganizing these helper functions into a separate file to be accessed when the course gql strings are used.
@Shrinjay could you address this comment?

frontend/src/APIClients/mutations/EntityMutations.ts
y2kwak: Yeah going forward probably not necessary. I wanted to apply the logic to Entities here for my understanding since I'm not able to test with Course components :))

frontend/src/APIClients/BaseAPIClient.ts
```diff
@@ -1,49 +1,113 @@
...
-      setLocalStorageObjProperty(
-        AUTHENTICATED_USER_KEY,
-        "accessToken",
-        accessToken,
-      );
-
-      newConfig.headers.Authorization = accessToken;
-    }
+const httpLink = (createUploadLink({
+  uri: `${process.env.REACT_APP_BACKEND_URL}/graphql`,
```
DinuWije, line 19: curious: do we not use react router for this?

frontend/src/APIClients/mutations/AuthMutations.ts
```diff
@@ -0,0 +1,57 @@
+import { gql } from "@apollo/client";
+
+export const LOGIN = gql`
+  mutation Login($email: String!, $password: String!) {
```
DinuWije, line 4: why is `login` a mutation instead of a query? are we changing stuff from a backend perspective?

frontend/src/App.tsx
```diff
@@ -50,59 +52,61 @@ const App = (): React.ReactElement => {
...
+          <AuthContext.Provider
+            value={{ authenticatedUser, setAuthenticatedUser }}
+          >
+            <ChakraProvider theme={defaultTheme}>
+              <Router>
+                <Switch>
+                  <Route exact path={Routes.LOGIN_PAGE} component={Login} />
+                  <Route exact path={Routes.SIGNUP_PAGE} component={Signup} />
+                  <PrivateRoute
+                    exact
```
DinuWije, line 65: not something we need to implement right now, but in the future if we wanted to make the homepage unauthenticated, is it just changing the route? also how are the different home pages for clients/admins expressed?

frontend/src/APIClients/queries/CourseQueries.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const COURSES = gql`
+  query Courses {
+    courses {
+      id
+      title
+      description
+      image
+      previewImage
+      modules
```
draymond63, line 11: image

I got this error when local hosting your branch. Please add the following to all the `module` attributes (including those in the mutations file)

```
    modules {
        id
        title
        description
        image
        previewImage
        published
        lessons
    }
```

y2kwak: fixed!

draymond63 (approved):
Works! There's a few things we will have to figure out (like the serialization accept) but I'm happy to merge this now :)

PR #107, opened 2022-03-27
Converting modules to an array, allow null response

draymond63:
## Implementation Description
There were a lot of files edited, but the main points are:
- Modules should now be a list of objects instead of a map
- Module Editor URL now uses the index of the module as the identifier for which module is being edited
- Response DTOs can now have null attributes to prevent GraphQL errors from occuring

## Screenshots
Same as before, not applicable.

- Should the URL of the module editor use the index of the array or still its ID?
- Make sure the admin dashboard still works!

## Checklist

### DIFF
--- a/backend/models/course.model.ts
+++ b/backend/models/course.model.ts
@@ -35,7 +35,7 @@ const ModuleSchema: Schema = new Schema({
     default: false,
   },
   lessons: [String],
-})
+});
 
 const CourseSchema: Schema = new Schema({
   title: {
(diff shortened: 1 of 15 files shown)


Shrinjay (approved):

backend/models/course.model.ts
```diff
@@ -25,6 +25,18 @@ export interface CourseVisibilityAttributes {
   published?: boolean;
 }
 
+const ModuleSchema: Schema = new Schema({
+  title: String,
```
molly-yu, line 29: I'm wondering if we should make the title required for the modules?

molly-yu (commented):
Aside from the linting and small comment I made above, I think this looks great! Thanks Daniel!

backend/models/course.model.ts
draymond63: Good call, I'll switch it

y2kwak (approved):

PR #111, opened 2022-03-31
Implement course/module visibility logic in course getters

e-wai:
## Implementation Description
- Reflect updated state of our documents (published attribute on modules, not courses) across models + interfaces + etc in backend
- Fix up course filtering for getCourse endpoint
- Filter through courses+modules for the following user/staff/admin access level logic:
```ts
case "User":
  return {
    includePrivateCourses: false,
    includeOnlyPublishedModules: true,
  };
case "Staff":
  return {
    includePrivateCourses: false,
    includeOnlyPublishedModules: false,
  };
case "Admin":
  return {
    includePrivateCourses: true,
    includeOnlyPublishedModules: false,
  };
```
- If `includedOnlyPublishedModules == true`, use aggregate query to unwrap module list, filter, and then return. Else use normal find query
  - Using aggregate seems to prefer the explicit `_id` type and doesn't find the id string alias

## Screenshots
Visual depiction of PR
n/a

- Access logic makes sense?
- MongoDB query makes sense + readable + not super inefficient?

## Testing Procedure

## Things to Note

## Checklist

Admin response:
```gql
{
  "data": {
    "courses": [
      {
        "id": "62454b6352e55a02773eb28d",
        "private": true,
        "modules": [
          {
            "published": true
          }
        ]
      },
      {
        "id": "62454b8052e55a02773eb28f",
        "private": false,
        "modules": [
          {
            "published": true
          }
        ]
      },
      {
        "id": "62454b9352e55a02773eb291",
        "private": true,
        "modules": [
          {
            "published": false
          }
        ]
      },
      {
        "id": "62454bd352e55a02773eb293",
        "private": false,
        "modules": []
      },
      {
        "id": "62454beb52e55a02773eb294",
        "private": false,
        "modules": [
          {
            "published": true
          },
          {
            "published": false
          }
        ]
      }
    ],
    "course": {
      "id": "62454beb52e55a02773eb294",
      "private": false,
      "modules": [
        {
          "published": true
        },
        {
          "published": false
        }
      ]
    }
  }
}
```

Staff response:
```gql
{
  "data": {
    "courses": [
      {
        "id": "62454b8052e55a02773eb28f",
        "private": false,
        "modules": [
          {
            "published": true
          }
        ]
      },
      {
        "id": "62454bd352e55a02773eb293",
        "private": false,
        "modules": []
      },
      {
        "id": "62454beb52e55a02773eb294",
        "private": false,
        "modules": [
          {
            "published": true
          },
          {
            "published": false
          }
        ]
      }
    ],
    "course": {
      "id": "62454beb52e55a02773eb294",
      "private": false,
      "modules": [
        {
          "published": true
        },
        {
          "published": false
        }
      ]
    }
  }
}
```

User:
```gql
{
  "data": {
    "courses": [
      {
        "id": "62454b8052e55a02773eb28f",
        "private": false,
        "modules": [
          {
            "published": true
          }
        ]
      },
      {
        "id": "62454bd352e55a02773eb293",
        "private": false,
        "modules": []
      },
      {
        "id": "62454beb52e55a02773eb294",
        "private": false,
        "modules": [
          {
            "published": true
          }
        ]
      }
    ],
    "course": {
      "id": "62454beb52e55a02773eb294",
      "private": false,
      "modules": [
        {
          "published": true
        }
      ]
    }
  }
}
```

### DIFF
--- a/backend/graphql/resolvers/courseResolvers.ts
+++ b/backend/graphql/resolvers/courseResolvers.ts
@@ -28,11 +28,20 @@ const getCourseVisibilityAttributes = (
 ): CourseVisibilityAttributes => {
   switch (role) {
     case "User":
-      return { private: false, published: true };
+      return {
+        includePrivateCourses: false,
+        includeOnlyPublishedModules: true,
+      };
     case "Staff":
-      return { published: true };
+      return {
+        includePrivateCourses: false,
+        includeOnlyPublishedModules: false,
+      };
     case "Admin":
-      return {};
+      return {
+        includePrivateCourses: true,
+        includeOnlyPublishedModules: false,
+      };
     default:
       return assertNever(role);
   }
--- a/backend/graphql/resolvers/courseResolvers.ts
+++ b/backend/graphql/resolvers/courseResolvers.ts
@@ -43,11 +43,22 @@ const courseResolvers = {
     course: async (
       _parent: undefined,
       { id }: { id: string },
+      context: ExpressContext,
     ): Promise<CourseResponseDTO> => {
-      return courseService.getCourse(id);
+      const accessToken = getAccessToken(context.req);
+      if (!accessToken) {
+        throw new AuthenticationError(
+          "Failed authentication and/or authorization by role",
+        );
+      }
+
+      const role = await authService.getUserRoleByAccessToken(accessToken);
+      const attributes = getCourseVisibilityAttributes(role);
+
+      return courseService.getCourse(id, attributes);
     },
     courses: async (
-      _parent: any,
+      _parent: undefined,
       _args: { [key: string]: any },
       context: ExpressContext,
     ): Promise<CourseResponseDTO[]> => {
--- a/backend/models/course.model.ts
+++ b/backend/models/course.model.ts
@@ -1,4 +1,4 @@
-import { Schema, Document, model } from "mongoose";
+import { Schema, Document, model, Types } from "mongoose";
 
 export interface Module {
   title: string;
@@ -10,7 +10,7 @@ export interface Module {
 }
 
 export interface Course extends Document {
-  id: string;
+  _id: Types.ObjectId;
   title: string;
   description: string;
   image: string;
--- a/backend/models/course.model.ts
+++ b/backend/models/course.model.ts
@@ -1,6 +1,7 @@
 import { Schema, Document, model, Types } from "mongoose";
 
 export interface Module {
+  _id: Types.ObjectId;
   title: string;
   description: string;
   image: string;
@@ -10,19 +11,18 @@ export interface Module {
 }
 
 export interface Course extends Document {
-  _id: Types.ObjectId;
+  id: string;
   title: string;
   description: string;
   image: string;
   previewImage: string;
   modules: Module[];
   private: boolean;
-  published: boolean;
 }
 
 export interface CourseVisibilityAttributes {
-  private?: boolean;
-  published?: boolean;
+  includePrivateCourses: boolean;
+  includeOnlyPublishedModules: boolean;
 }
 
 const ModuleSchema: Schema = new Schema({
@@ -61,11 +61,6 @@ const CourseSchema: Schema = new Schema({
   private: {
     type: Boolean,
     default: true,
-    required: true,
-  },
-  published: {
-    type: Boolean,
-    default: false,
     required: true,
   },
 });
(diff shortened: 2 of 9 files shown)


backend/services/implementations/courseService.ts
```diff
@@ -13,22 +14,58 @@ import MgCourse, {
...
+  return filterForPublished
+    ? [
+        {
+          $project: {
+            title: true,
+            description: true,
+            image: true,
+            previewImage: true,
+            private: true,
+            published: true,
```
draymond63, line 27: We should be removing the `published` attribute on the course document, right?

backend/models/course.model.ts
```diff
@@ -10,7 +10,7 @@ export interface Module {
 }
 
 export interface Course extends Document {
-  id: string;
+  _id: Types.ObjectId;
```
draymond63, line 13: Good addition, but does the Document interface not already have the `_id` attribute defined?

draymond63, line 13: Good addition, but does the Document interface not already have the `_id` attribute defined?

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -43,11 +43,22 @@ const courseResolvers = {
...
-      return courseService.getCourse(id);
+      const accessToken = getAccessToken(context.req);
+      if (!accessToken) {
+        throw new AuthenticationError(
+          "Failed authentication and/or authorization by role",
+        );
+      }
+
+      const role = await authService.getUserRoleByAccessToken(accessToken);
+      const attributes = getCourseVisibilityAttributes(role);
```
draymond63, line 56: Since courses will no longer have a published attribute (that's only for courses), this is a bit out of date no?

backend/services/implementations/courseService.ts
```diff
@@ -13,22 +14,58 @@ import MgCourse, {
...
+
+      [course] = courseQueryResult;
     } catch (error: unknown) {
       Logger.error(`Failed to get course. Reason = ${getErrorMessage(error)}`);
       throw error;
     }
 
     return {
-      id: course.id,
+      id: course._id.toHexString(),
```
draymond63, line 68: Linter isn't happy with this :eyes:

e-wai: Cheated this

backend/models/course.model.ts
e-wai: Yeah seems fine without

e-wai (commented):
@draymond63 Remaining thoughts:
- Does this affect the updateCourse query?
- I only handled backend here, does this break the frontend?

backend/services/implementations/courseService.ts
```diff
@@ -13,47 +14,105 @@ import MgCourse, {
...
+          [{ $match: queryAttributes }, ...publishedModulesOnlyQuery],
+        ).exec();
+
+        if (!courseQueryResult || courseQueryResult.length === 0) {
+          throw new Error(`Course id ${id} not found`);
+        }
+
+        [course] = courseQueryResult;
+      } else {
+        course = await MgCourse.findOne(queryAttributes);
```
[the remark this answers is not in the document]
draymond63: I think I like this solution as it's easier to read. I'm okay with using findOne even though we are searching by id.

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -28,11 +28,20 @@ const getCourseVisibilityAttributes = (
 ): CourseVisibilityAttributes => {
   switch (role) {
     case "User":
-      return { private: false, published: true };
+      return {
+        includePrivateCourses: false,
+        includeOnlyPublishedModules: true,
```
draymond63, line 33: Maybe it could be worth chaning it to `includeDraftModules`? I think that makes more sense, as long as it's clear that draft is the opposite of published.

draymond63 (approved):
I realize there's still some open questions, but I'm happy with the way this is being implemented. We will discuss at tomorrow's meeting, but I think things are mostly good!

backend/graphql/resolvers/courseResolvers.ts
e-wai: Had this discussion synchronously and both had weak opinions, but basically:
- Idea of "draft" is currently limited to frontend, don't see a compelling argument to bring it into the backend esp. since the backend calls it published and therefore anyone working in the backend (probably) understands it easiest this way
- The double negative is gross but not to the point that it's not understandable
- Can always adjust it as we go, as other devs touch this code

@draymond63 did i miss anything?

PR #113, opened 2022-04-03
Module Editor Sidebar Skeleton

y2kwak:
## Implementation Description
Setting up skeleton for module editor sidebar!!

## Screenshots
Screen Shot 2022-04-21 at 21 38 10

- 

## Testing Procedure
- Test on different screen sizes for responsiveness
- Compare with[ figma designs](https://www.figma.com/file/Dimm13cjrpYU5LkcrV6VxF/Course-Editor?node-id=691%3A3296)

## Things to Note

## Checklist


frontend/src/components/module-editor/SideBar.tsx
```diff
@@ -1,46 +1,126 @@
...
 const Sidebar = (): React.ReactElement => {
+  const { moduleIndex }: ModuleEditorParams = useParams();
+
+  const context = useContext(EditorContext);
+
+  if (!context) return <></>;
+  const { state } = context;
   return (
-    <Box w="20%">
+    <Box w="20%" minW="300px">
```
draymond63, line 33: not a big fan of setting the min width to a pixel value. Does it not work with `min-content` or `max-content`?

frontend/src/components/module-editor/SideBar.tsx
```diff
@@ -1,46 +1,126 @@
...
       <Flex
         position="fixed"
-        top="0"
-        left="0"
-        p="4"
         w="inherit"
-        h="100%"
-        minW="min-content"
+        minW="300px"
```
draymond63, line 37: If the parent has a minimum width, this flex should simply receive it's width right? Or am I wrong on that? There might be something weird about using `min-content` when the child is of `position=fixed`. Anyway, if it works it works (but I'm still not a fan of the pixel value)

draymond63 (approved):
Happy with the implementation! Not sure why the preview deploy didn't work, but I'm not going to worry about it for now. A few comments about the sizing of the sidebar, but I'd be happy to merge it if you are :)

PR #114, opened 2022-04-03
Building Sign Up Page

binalpreetkalra:
## Implementation Description
Built the signup page flow with error states for each required input and password validation. Checks if password meets the 6 character requirement and matches the re-entered password for further confirmation.

## Screenshots
image
image
image

- Ensuring the signup page format is adequate on their desktop screen
- The back to sign-in page link works
- Unable to submit form unless every field is inputted and password meets the requirements + match
- Error messages as needed show up

## Testing Procedure

## Next Steps
- Have the main image on the right change size as needed on different screens
- Ensure the error messages are adequate for the user (maybe add more detail)
- Fix spacing for the error messages above the Sign Up button
- Show loading icon after user logs in as the next page is being loaded

## Checklist


draymond63 (approved):
There are some things that would be great to add, but I think it's good to merge! Thanks so much for your work this term Binal!

PR #116, opened 2022-04-10
Setup manage users admin page

e-wai:
## Implementation Description
This PR will setup the manage users page

[modified]
- make users and course separate routes (so that when we navigate back and forward, it includes these pages separately in the history)
- delete the `AdminDashboard` combined component and move to separate components in `ManageUsers` and `ManageCourses` pages
- change sidebar to  have tabs be composed of buttons

[new]
- Create `UserCard` component to render a spotlighted user's information. Add equivalent `NoUserCard` if no spotlighted user
- Create `SearchUsersBar` component for searching for user by email -- EXACT MATCH ONLY
- Create rest of users page to match Figma designs here -- https://www.figma.com/file/Yb3MvhfQAhfCjLg8cnNCoo/User-Management?node-id=112%3A61
  - Uses query for all users (and sets up relevant API things). Caches on user side which produces an edge case if page is not refreshed for significant time while user goes through sign up flow, reasonable mitigation by forcing query refresh within ~5 mins?

## Screenshots

https://user-images.githubusercontent.com/40974372/162696816-410c916f-f5a3-422f-a74b-f72879e1561f.mov

- Better to use another component instead of <Menu>?

*Outstanding work/questions*
- Order of fields is wrong, but are we ok with that
- should prob have a redirect if they go to `/admin`
- Make admin pages such as this one only accessible to admin users

https://user-images.githubusercontent.com/40974372/162700457-07e36dc3-0274-49b9-9f15-7b49948877d3.mov

- The search bar has some really weird behaviour -> on correct answer typed, loses focus on the input field. But this doesn't happen when a correct email is copied exactly

https://user-images.githubusercontent.com/40974372/162699022-72e7c6a1-56c3-4d9f-ae2b-bafb2cb3d68f.mov

- because the list loses focus when the input field is clicked, the result disappears and you have to delete/type it for it to reappear

## Testing Procedure
   - Is it secure? Could nefarious users access/break it? *be careful since all users information loaded in frontend, but not PII*

## Things to Note
- Additional dependencies/libraries you've added? => None
  - *Check the users query edge case (see top section of summary of implementation)

## Checklist


frontend/src/theme/components/button.ts
```diff
@@ -121,13 +121,6 @@ const Button = {
       fontSize: "xs",
       height: "32px",
     },
-    transparent: {
-      bg: "transparent",
-      padding: "0px",
-      margin: "0px",
-      border: "0px",
-      fontWeight: "normal",
-    },
```
e-wai, line 130: unused + not design system

frontend/src/components/pages/ManageUsersPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [users, setUsers] = useState<UsersByEmail>({});
+
+  useQuery<{
+    users: Array<UserResponse>;
+  }>(USERS, {
+    fetchPolicy: "cache-and-network",
+    onCompleted: (data) => {
+      const userProfiles = data.users.reduce(
+        (map: UsersByEmail, user: UserResponse) => {
+          // eslint-disable-next-line no-param-reassign
```
e-wai, line 50: we've used this workaround before, couldn't quite find what the expectation is here

frontend/src/components/pages/ManageUsersPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Flex
+          my={6}
+          // px instead of mx to extend border completely in container
+          px={9}
+          pb={6}
+          justify="space-between"
+          borderBottom="1px"
+          borderColor="background.lightgrey"
+        >
+          <HStack>
```
e-wai, line 74: Allows the space-between property to exist between (title + bar) and (download button) rather than (title + bar + download button). Flex property doesn't exist on `SearchUsersBar` so I thought this might be easiest if not the prettiest

frontend/src/components/admin/SearchUsersBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+          placeholder="Search users by email"
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+            setSearchEmail(event.target.value);
+            setShowNewResult(true);
+          }}
+          onFocus={() => setShowNewResult(true)}
+          onKeyPress={(event: React.KeyboardEvent<HTMLInputElement>) => {
+            if (event.key === "Enter" && searchEmail in users) {
+              handleOnSelectUser();
+            }
```
e-wai, line 45: 100% recommend testing this out yourself in staging/locally to see this interaction (both through clicks and using keyboard)

frontend/src/components/pages/ManageUsersPage.tsx
draymond63: Yep, definitely should just disable the linter here. Only other solutions I see are to use object spreading, but that would create a new object for every call of the function which isn't exactly performant.

frontend/src/constants/Routes.ts
```diff
@@ -14,6 +14,8 @@ export const UPDATE_ENTITY_PAGE = "/entity/update";
 
 export const HOOKS_PAGE = "/hooks";
 
-export const ADMIN_DASHBOARD_PAGE = "/admin";
```
draymond63, line 17: As you mentioned, lets add a redirect for this url. Sending to `/admin/courses` I think is best, as I believe they would use it more frequently.

frontend/src/components/pages/ManageUsersPage.tsx
```diff
@@ -0,0 +1,108 @@
...
+    spotlightedUser,
+    setSpotlightedUser,
+  ] = useState<UserCardProps | null>();
+
+  // Never expose this
+  const [users, setUsers] = useState<UsersByEmail>({});
+
+  useQuery<{
+    users: Array<UserResponse>;
+  }>(USERS, {
```
draymond63, line 45: So we are retrieving all the users on load, and then searching through the response when the user types? Does this mean all of the search logic is happening on the frontend?

Not the biggest fan of this, because even though this is a protected route:
1. The admin will have access to all user emails if they try (not a big at all, but idealogically maybe bad)
2. doesn't this mean we have to read all the mongo db users whenever this page loads? Seems impractical

Anyway, this is all to say, why aren't we using the `getUserByEmail` endpoint? This API seems perfect for this application, unless I'm misunderstanding something. That way, it's more efficient, and all of the backend/search functionality can be in the `SearchUsersBar` component

draymond63 (changes requested):
Lots of progress!

Only two things I'd say are keeping this from being merged.
1. I think we should use the `` query instead of retrieving everbody with the `users` query.
2. When I type in the searchbar on localhost, I receive the following warning in the console.
 
image
Not a big deal, but it does clutter up the console which is annoying while debugging. Not sure if it was just me experiencing this.

Anyway, happy to discuss these comments. Great work!

frontend/src/components/pages/ManageUsersPage.tsx
e-wai: Yes that's exactly what's happening right now. It doesn't feel like good practice -- and probably isn't -- but at least for the points you mentioned I still think it makes more sense:
- the admin will have access to all user emails no matter if the search in the frontend or the backend the choice of where we're storing everything does not change access at all
  - to me, that's something that's reasonably within the admin's capabilities
- yes, we will have to read all mongo db users whenever the page loads and I agree it's not the most practical, but isn't using the `getUserByEmail` query even worse? You'd be submitting an API call on every keystroke 
  - even if we sub this with something debounced on 50ms for eg., you'd be submitting a request every 50ms still, which is more frequent than page refreshes
  - pretty sure the `getUserByEmail` query would read through all the documents too? not saving us anything there
 
Is there another solution that I'm missing here?

e-wai:
Not convinced the margin warning is me -- I don't apply margin properties in that file, only offset?

frontend/src/components/pages/ManageUsersPage.tsx
e-wai: OK discussed at work session, satisfied to move to `getUserByEmail`

draymond63 (approved):
Looks good! The only weird issue is losing focus on a match, making it so you can't press enter. But! We can add to the niceties list.

e-wai:
yeahh, for some reason when a user is found then it defocuses on the input bar. When the MenuItem is set to `isDisabled` instead on having an `onClick`, it doesn't defocus? not sure but something to play around with later

PR #122, opened 2022-05-06
Update account default roles depending on the environment

jfdoming:
## Implementation Description
In production, default accounts to the "User" role. Otherwise, default accounts to the "Admin" role.

## Screenshots
N/A

- Is this implementation secure enough?

## Testing Procedure

Test via creating a new user locally, then verifying that user has admin permissions.

## Things to Note

## Checklist


draymond63 (approved):
Sweet!

jfdoming:
Created an issue to track reverting this change. Merging.

PR #128, opened 2022-05-10
Function New Lesson Saving

draymond63:
## Implementation Description
Brief summary of PR

## Screenshots
Visual depiction of PR

- 

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/components/module-editor/SideBar.tsx
+++ b/frontend/src/components/module-editor/SideBar.tsx
@@ -67,8 +67,6 @@ const Sidebar = (): React.ReactElement => {
     });
     if (!data) throw Error("Failed to create new lesson");
     const lesson = data.createLesson;
-    // Make lesson response compatible with what the frontend needs
-    // const newLesson = formatLessonResponse(lesson);
     const { id: newID } = lesson;
     // Update lesson ID in state with response
     dispatch({ type: "update-lesson-id", value: { oldID, newID } });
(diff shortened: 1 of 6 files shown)


frontend/src/reducers/ModuleEditorContextReducer.ts
```diff
@@ -6,10 +6,21 @@ import {
...
 } from "../types/ModuleEditorTypes";
 
 /* eslint-disable no-console */
 
+const updateChangeStatus = (
+  hasChanged: EditorChangeStatuses,
+  docID: string,
+  status: EditorChangeStatus,
+): EditorChangeStatuses => {
+  if (!(docID in hasChanged)) return { ...hasChanged, [docID]: status };
```
jfdoming, line 20: I believe this logic breaks for DELETE actions; if the user performs a CREATE or UPDATE first, then a DELETE, the final state of the system should be that the document is deleted, but based on this logic, that won't happen. I think we probably need some special cases here; maybe we can leave a TODO for when we add the ability to delete lessons individually?

frontend/src/components/module-editor/SideBar.tsx
```diff
@@ -43,47 +44,60 @@ const Sidebar = (): React.ReactElement => {
...
+    changedLesson: LessonRequest,
+  ) => {
+    const { data } = await createLesson({
       variables: { lesson: changedLesson },
     });
-    // TODO: Update lesson ID in state with response
+    if (!data) throw Error("Failed to create new lesson");
+    const lesson = data.createLesson;
+    // Make lesson response compatible with what the frontend needs
+    // const newLesson = formatLessonResponse(lesson);
```
jfdoming, line 71: Is this line intended to be left in?

jfdoming (commented):
Good changes! Left some minor comments.

PR #141, opened 2022-05-19
Add edit/copy/delete buttons on content block hover

e-wai:
Note: I'm still unhappy with the padding on the left side (when you hover, you can see that the divider doesn't extend fully to the left, because of the sidebar). Not blocking and high-priority, but little fix TODO. Similar for padding between buttons, it seems to force a width on the icon button

## Implementation Description
- Add icon buttons to each content block, for edit, copy, and delete options
- Dummy buttons right now (not doing anything)

## Screenshots

https://user-images.githubusercontent.com/40974372/169189856-4cd88946-f7de-4ce6-a846-b8a6ad435c28.mov

Compare to [figma](https://www.figma.com/file/Dimm13cjrpYU5LkcrV6VxF/Course-Editor?node-id=691%3A3296):

Screen Shot 2022-05-18 at 10 25 02 PM
- slight changes to the icons used (used the default Chakra icons)
- check spacing? I matched the Figma but will double check because it looks different

- What's the right approach for adding those "on click" functions for edit / copy / delete? I assumed that we want the funcs to be sent in as props, that way it can edit the "unsaved changes" in the parent component
- Accessibility-wise, is it the right tab index approach? I kept the component as existing with zero opacity (matching the drag+drop operator on left side) so that spacing was same no matter the hover state,  but it means I needed to override the tab index manually

## Testing Procedure
- What test(s) should we run to make sure your code works? manual QA

## Things to Note
- Additional dependencies/libraries you've added? nope

## Checklist

### DIFF
--- a/frontend/src/components/module-editor/EditContentOptionsMenu.tsx
+++ b/frontend/src/components/module-editor/EditContentOptionsMenu.tsx
@@ -1,0 +1,39 @@
+import React from "react";
+import { HStack, IconButton } from "@chakra-ui/react";
+import { EditIcon, CopyIcon, DeleteIcon } from "@chakra-ui/icons";
+import { EditContentOptionsMenuProps } from "../../types/ModuleEditorTypes";
+
+const EditContentOptionsMenu = ({
+  isVisible,
+  onEditClick,
+  onCopyClick,
+  onDeleteClick,
+}: EditContentOptionsMenuProps): React.ReactElement => {
+  return (
+    <HStack spacing={0} visibility={isVisible ? "visible" : "hidden"}>
+      <IconButton
+        aria-label="Edit this content block"
+        variant="transparent"
+        onClick={onEditClick}
+        m={0}
+        icon={<EditIcon color="background.dark" h={5} w={5} />}
+      />
+      <IconButton
+        aria-label="Duplicate this content block"
+        variant="transparent"
+        onClick={onCopyClick}
+        m={0}
+        icon={<CopyIcon color="background.dark" h={5} w={5} />}
+      />
+      <IconButton
+        aria-label="Delete this content block from lesson"
+        variant="transparent"
+        onClick={onDeleteClick}
+        m={0}
+        icon={<DeleteIcon color="background.dark" h={5} w={5} />}
+      />
+    </HStack>
+  );
+};
+
+export default EditContentOptionsMenu;
--- a/frontend/src/components/module-editor/EditableContentBlock.tsx
+++ b/frontend/src/components/module-editor/EditableContentBlock.tsx
@@ -4,6 +4,7 @@ import { Draggable } from "react-beautiful-dnd";
 
 import { ContentBlock, ContentTypeEnum } from "../../types/ModuleEditorTypes";
 import { TextBlock, ImageBlock } from "../common/content";
+import EditContentOptionsMenu from "./EditContentOptionsMenu";
 
 /* eslint-disable react/jsx-props-no-spreading */
 
@@ -45,7 +46,7 @@ const EditableContentBlock = ({
           style={provided.draggableProps.style}
         >
           <Divider opacity={isHovered ? 1 : 0} />
-          <Flex justify="space-between">
+          <Flex width="100%" justify="space-between">
             <Box opacity={isHovered ? 1 : 0} {...provided.dragHandleProps}>
               <svg width="24" height="24" viewBox="0 0 24 24">
                 <path
@@ -55,6 +56,12 @@ const EditableContentBlock = ({
               </svg>
             </Box>
             {SelectContentBlock(block)}
+            <EditContentOptionsMenu
+              isVisible={isHovered}
+              onEditClick={() => {}}
+              onCopyClick={() => {}}
+              onDeleteClick={() => {}}
+            />
           </Flex>
           <Divider opacity={isHovered ? 1 : 0} />
         </VStack>
--- a/frontend/src/theme/components/button.ts
+++ b/frontend/src/theme/components/button.ts
@@ -121,6 +121,13 @@ const Button = {
       fontSize: "xs",
       height: "32px",
     },
+    transparent: {
+      bg: "transparent",
+      padding: "0px",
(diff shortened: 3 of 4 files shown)


frontend/src/theme/components/button.ts
```diff
@@ -121,6 +121,13 @@ const Button = {
       fontSize: "xs",
       height: "32px",
     },
+    transparent: {
+      bg: "transparent",
+      padding: "0px",
+      margin: "0px",
+      border: "0px",
+      fontWeight: "normal",
+    },
```
e-wai, line 130: Basically used this as a way to avoid it going to default styling. Not sure if there's something similar that already exists?

frontend/src/components/module-editor/EditableContentBlock.tsx
```diff
@@ -37,6 +38,7 @@ const EditableContentBlock = ({
         <VStack
           width="100%"
           padding="0.5rem 1rem"
+          // pv={2}
```
e-wai, line 41: shoot, will clean this up

frontend/src/components/module-editor/EditableContentBlock.tsx
```diff
@@ -55,6 +57,12 @@ const EditableContentBlock = ({
               </svg>
             </Box>
             {SelectContentBlock(block)}
+            <EditContentOptionsMenu
+              isVisible={isHovered}
```
jfdoming, line 61: Rather than passing in a prop, is there any reason not to just conditionally render this? i.e.
```suggestion
            {isHovered && <EditContentOptionsMenu
            ...
```

jfdoming, line 61: Rather than passing in a prop, is there any reason not to just conditionally render this? i.e.
```suggestion
            {isHovered && <EditContentOptionsMenu
            ...
```

frontend/src/components/module-editor/EditContentOptionsMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <HStack spacing={4} opacity={isVisible ? 1 : 0}>
+      <IconButton
+        aria-label="Edit this content block"
+        tabIndex={currentTabIndex}
+        variant="transparent"
+        onClick={onEditClick}
+        m={0}
+        icon={<EditIcon color="background.dark" h={5} w={5} />}
+      />
```
jfdoming, line 23: (optional) We could extract this into a separate component potentially:
```suggestion
      <TransparentButton
        label="Edit this content block"
        tabIndex={currentTabIndex}
        onClick={onEditClick}
        icon={EditIcon}
      />
```

jfdoming, line 23: (optional) We could extract this into a separate component potentially:
```suggestion
      <TransparentButton
        label="Edit this content block"
        tabIndex={currentTabIndex}
        onClick={onEditClick}
        icon={EditIcon}
      />
```

frontend/src/components/module-editor/EditContentOptionsMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+  onCopyClick,
+  onDeleteClick,
+}: EditContentOptionsMenuProps): React.ReactElement => {
+  const currentTabIndex = isVisible ? 0 : -1;
+
+  return (
+    <HStack spacing={4} opacity={isVisible ? 1 : 0}>
+      <IconButton
+        aria-label="Edit this content block"
+        tabIndex={currentTabIndex}
```
jfdoming, line 18: If we do the conditional rendering, we might also be able to avoid having to set the `tabIndex` manually.

jfdoming, line 18: If we do the conditional rendering, we might also be able to avoid having to set the `tabIndex` manually.

frontend/src/components/module-editor/EditableContentBlock.tsx
```diff
@@ -55,6 +57,12 @@ const EditableContentBlock = ({
...
+              onEditClick={() => console.log("edit")}
+              onCopyClick={() => console.log("copy")}
+              onDeleteClick={() => console.log("delete")}
```
jfdoming, line 64: (nit, optional): We could change these to empty functions for now, since otherwise these warnings will show up on other people's PRs until the subsequent ticket is worked on.
```suggestion
              onDeleteClick={() => {}}
```

jfdoming, line 64: (nit, optional): We could change these to empty functions for now, since otherwise these warnings will show up on other people's PRs until the subsequent ticket is worked on.
```suggestion
              onDeleteClick={() => {}}
```

jfdoming (changes requested):
One question about conditional rendering, but excellent job!

frontend/src/components/module-editor/EditContentOptionsMenu.tsx
e-wai: Discussed sync -- conditional rendering is not great for us since we want the text container to stay in the same place, no matter if we're rendering the options menu (aka when hovered). Instead, switch to using `visibility` attribute which will handle the tab index + opacity all in one go

frontend/src/components/module-editor/EditContentOptionsMenu.tsx
e-wai: Hmm I may leave it for now - I could see it cleaning up within this file, I'm not sure creating a transparent button with these stylings has much application (that isn't already handled by the new `button.ts` styling) otherwise

e-wai (commented):
Updated look

https://user-images.githubusercontent.com/40974372/169931480-c6c53311-d82e-40e8-b189-86a9f3c39358.mov

frontend/src/components/module-editor/EditableContentBlock.tsx
e-wai: (on comment below)

jfdoming (approved):
LGTM!

PR #143, opened 2022-05-24
Implement module editor component selector options

y2kwak:
## Implementation Description
- Adding new content types
- Adding content icons in assets folder

## Screenshots

Screen Shot 2022-05-23 at 22 58 03

- Is there a better way to implement [icon switch case](https://github.com/uwblueprint/rowan-house/pull/143/files#diff-fc330b5dbcacbd5aeaa46a44589f686d0362410d8b0cced6d47dd9c191547c30R23-R49) each content icon to avoid switch cases?

## Testing Procedure

## Things to Note

## Checklist


frontend/src/components/module-editor/BlockPreview.tsx
```diff
@@ -13,14 +20,48 @@ const BlockStyle = ({
   className?: string;
   content: ContentTypeEnum;
 }): React.ReactElement => {
+  let contentIcon: React.ReactElement = <ImageIcon />;
+  // eslint-disable-next-line default-case
```
mslwang, line 24: I know you likely won't be hitting the default case on an enum-based switch statement, but it's probably best to avoid ignoring the eslint error here and adding a default case.

frontend/src/components/module-editor/BlockPreview.tsx
```diff
@@ -13,14 +20,50 @@ const BlockStyle = ({
   className?: string;
   content: ContentTypeEnum;
 }): React.ReactElement => {
+  let contentIcon: React.ReactElement;
```
jfdoming, line 23: Honestly, the switch statement is fine, but since you mentioned in your PR, one other option might be:

```javascript
const ICONS = {
    "column.svg": ColumnIcon,
    ...
};

// later
if (Object.keys(ICONS).includes(content.preview)) {
    const IconComponent = ICONS[content.preview];
    contentIcon = <IconComponent width="fit-content" height="fit-content" />;
} else {
    contentIcon = <ImageIcon />;
}
```

jfdoming (approved):
LGTM. One bit of feedback based on the requested part of the code.

frontend/src/components/module-editor/BlockPreview.tsx
y2kwak: Hm, will leave it as switch cases for now then. Maybe if we end up adding more content types we'll need to refactor accordingly

PR #147, opened 2022-05-26
[RHS-145]: Fix user access to list courses endpoint

jfdoming:
## Implementation Description
Details: Depending on whether a user is a User or an Admin, we fetch courses from the DB in different ways. Either we fetch them using find (Admins) or using aggregate (Users). However, results from an aggregation are missing the .id field by default, which causes GraphQL queries which expect that field to fail (see [this issue](https://github.com/Automattic/mongoose/issues/8784)).  To fix this, we can pass the aggregation results through MgCourse.hydrate, which will populate the expected fields.

## Screenshots
N/A

- 

## Testing Procedure
- What test(s) should we run to make sure your code works? Set a user to the "User" role in Mongo, then load the admin homepage `/admin/courses`. If there's no crash, this fix is working.
   - What ways could the input be wrong? N/A
   - Is it secure? Could nefarious users access/break it? See below.

## Things to Note
The fact that "User" users can see the admin page might be seen as a security issue, but since we're packaging both admin and non-admin routes in the same bundle, a nefarious user could see these pages regardless. (Also, the relevant backend routes are already protected by RBAC.) We might consider (in a future ticket) redirecting "User" users away from this page, however, since otherwise they might encounter a worse UX (when random portions of the page don't work due to the backend denying their permission).

## Checklist

### DIFF
--- a/backend/services/implementations/courseService.ts
+++ b/backend/services/implementations/courseService.ts
@@ -51,11 +51,9 @@ class CourseService implements ICourseService {
           [{ $match: queryAttributes }, ...publishedModulesOnlyQuery],
         ).exec();
 
-        if (!courseQueryResult || courseQueryResult.length === 0) {
-          throw new Error(`Course id ${id} not found`);
-        }
-
-        [course] = courseQueryResult;
+        course = courseQueryResult?.length
+          ? MgCourse.hydrate(courseQueryResult[0])
+          : null;
       } else {
         course = await MgCourse.findOne(queryAttributes);
       }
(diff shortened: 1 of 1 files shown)


mslwang:
Quick question, why are we using aggregation instead of fetching for Users? Briefly looking at the call, I feel like something similar can be implemented using find since it seems like a filter-only call (no reduction/aggregation). It's not a huge deal, but just curious.

jfdoming:
> Quick question, why are we using aggregation instead of fetching for Users? Briefly looking at the call, I feel like something similar can be implemented using find since it seems like a filter-only call (no reduction/aggregation). It's not a huge deal, but just curious.

Great question! The issue is that we need to filter on the subdocument (modules), and actually modify the returned document (we don't want to return ALL the modules in a particular course, only the visible ones). Definitely a good instinct to stick to the simpler filter though.

draymond63 (approved):
Change seems innocent enough, I think it makes sense. Someone else should approve before merging.

PR #159, opened 2022-06-02
[RHS-158]: Add linter pre-commit hooks

jfdoming:
## Implementation Description
- Hint at devs to install pre-commit on docker startup
- Lint on pre-commit
- Use linter cache to speed up pre-commit hooks

## Screenshots
image

- Can you think of cleaner/faster ways to implement this?

## Testing Procedure
A: Not designed to be secure, only hint to install pre-commit hooks for users.

## Checklist
- [x] Ensure code follows style guide by running the linters (:D)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 **/*.cache
 **/*.egg-info
 .history/
+.eslintcache
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+repos:
+-   repo: local
+    hooks:
+    -   id: felint
+        name: Lint frontend files
+        files: ^frontend/.*\.[jt]sx?$
+        types: [file]
+        language: system
+        entry: ./scripts/lint.sh felint
+    -   id: belint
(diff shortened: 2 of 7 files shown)


mslwang:
Would it be possible to try to use [husky](https://www.npmjs.com/package/husky) for pre-commits? It integrates naturally into a js/ts project, so we might be able to include it as part of installs somewhere.

I've been using a vscode extension for watching files, but I like the idea of using pre-commit hooks since it happens all at once.

e-wai (commented):
Says there's a symbolic link at `utils.sh` now, what does that do for us?

jfdoming:
> Says there's a symbolic link at `utils.sh` now, what does that do for us?

In theory it's not necessary, but I didn't want to break people's local setups in case they were using `utils.sh` in their own environment configuration.

mslwang:
Would it be possible to get instructions on how to set this up? I think it'd be great to get this pushed out ASAP so we can start linting better sooner! We can definitely consider husky later.

mslwang (approved):

PR #160, opened 2022-06-02
[RHS-130]: Add email verification workflow and authentication

mslwang:
## Implementation Description
After signing up, users that have not verified should not be able to access any private pages until they have verified their email. As such, we have created a page after sign up that reminds users to verify their email and all PrivateRoutes redirect to this page until they have verified. Furthermore, we have appended an email verification status property to the now termed AuthUser, adding backend support for this property.

## Screenshots
image

- See if it works correctly
- Is the user flow intuitive?

## Testing Procedure
- Go through signup flow and verify your email, then reload page to see if it works
- Try going to different pages after going through the signup flow to see what you can reach while "logged in" but not verified

## Things to Note
What should be the user flow for people who signup but want to leave the page, then try to return to login later? This is avoided when you let them in without having them verify. One thought I have is logging out as soon as you leave the verify email page without verifying your email, but a discussion may need to be had about this.

## Checklist


frontend/src/types/AuthTypes.ts
```diff
@@ -1,10 +1,11 @@
-export type AuthenticatedUser = {
```
jfdoming, line 1: Is there a particular reason to rename in this way? It makes the diff pretty large and it would be nice to keep the unshortened version unless there's a concrete benefit to doing so.

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,44 @@
...
+import authAPIClient from "../../APIClients/AuthAPIClient";
+
+const VerifyEmail = (): React.ReactElement => {
+  const { authUser, setAuthUser } = useContext(AuthContext);
+  const { refetch } = useQuery<
+    { userWithVerificationStatusByEmail: Omit<AuthUser, "accessToken"> },
+    { email: string }
+  >(GET_USER_WITH_VERIFICATION_STATUS_BY_EMAIL, { skip: true });
+  useEffect(() => {
+    if (authUser && authUser.email) {
```
jfdoming, line 19:
```suggestion
    if (authUser?.email) {
```

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,44 @@
...
+      authAPIClient.getAuthUser(authUser, refetch).then((user) => {
+        setAuthUser(user);
+      });
```
jfdoming, line 22: Does this work?
```suggestion
      authAPIClient.getAuthUser(authUser, refetch).then(setAuthUser);
```

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }, [authUser, setAuthUser, refetch]);
+
+  if (authUser && authUser?.emailVerified) {
```
jfdoming, line 26:
```suggestion
  if (authUser?.emailVerified) {
```

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { MANAGE_COURSES_PAGE } from "../../constants/Routes";
+import { AuthUser } from "../../types/AuthTypes";
+import authAPIClient from "../../APIClients/AuthAPIClient";
+
+const VerifyEmail = (): React.ReactElement => {
+  const { authUser, setAuthUser } = useContext(AuthContext);
+  const { refetch } = useQuery<
+    { userWithVerificationStatusByEmail: Omit<AuthUser, "accessToken"> },
+    { email: string }
+  >(GET_USER_WITH_VERIFICATION_STATUS_BY_EMAIL, { skip: true });
```
jfdoming, line 17: Maybe add a comment about why we have this line for people who aren't familiar with what `skip: true` does.

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <Flex minH="100vh">
+      <Center flex="1">
+        <VStack>
+          <Image height="13vh" marginBottom="2.5vh" src={RHSLogo} />
+          <Text variant="display-md" paddingBottom="1vw">
+            Please Verify your Email to Continue
+          </Text>
+        </VStack>
+      </Center>
+    </Flex>
```
jfdoming, line 40: Optional: I'm assuming this is identical to the code on the login page. What does the feasibility of refactoring this to a reusable wrapper component look like? e.g.
```suggestion
  <AuthPage>
    <Text variant="display-md" paddingBottom="1vw">
      Please Verify your Email to Continue
    </Text>
  </AuthPage>
```

jfdoming (changes requested):
Got some comments but overall great work!

frontend/src/types/AuthTypes.ts
mslwang: I understand that, it's just that it's a misnomer to say that this user will be necessarily authenticated because they may not be verified. Therefore, I think AuthUser is a better name for this. I can understand if you don't think it's worth it though.

frontend/src/components/pages/VerifyEmail.tsx
mslwang: We definitely can, but I don't know if AuthPage is the right name for it tbh. It's good to have it for the sake of consistency even if it's not much.

frontend/src/types/AuthTypes.ts
jfdoming: Let's keep it the same if possible, if you don't mind reverting everything back? If you feel strongly about it we can make a separate ticket and PR for that change.

frontend/src/components/pages/VerifyEmail.tsx
jfdoming: Yeah, AuthPage was just me spitballing. I'll make a separate ticket for this.

backend/services/implementations/userService.ts
```diff
@@ -113,9 +121,18 @@ class UserService implements IUserService {
...
   }
 
+  async getUserByEmail(email: string): Promise<UserDTO> {
+    const {
+      emailVerified,
+      ...rest
+    } = await this.getUserWithVerificationStatusByEmail(email);
+    return rest;
+  }
+
```
e-wai, line 135: Why not just combine it into a single endpoint for user + user email verified? Is there a situation where we want one but not the other?

frontend/src/types/AuthTypes.ts
e-wai: I'd also argue "authenticated" doesn't mean the same as "verified"? 

(If they've logged in successfully, they're still authenticated. Unverified email is a separate concept in my mind)

e-wai:
> What should be the user flow for people who signup but want to leave the page, then try to return to login later? This is avoided when you let them in without having them verify. One thought I have is logging out as soon as you leave the verify email page without verifying your email, but a discussion may need to be had about this.

Was there a conclusion on this? Maybe becomes a nitpicky product question, but I think a better user experience would be limiting functionality for accounts with unverified emails, without a hard block on them looking around. Admittedly that might be a lot more dev work, but I don't think the whole "nothing here" screen is an especially welcoming experience for a new user that could just be looking to poke around

backend/services/implementations/userService.ts
mslwang: Hmmm, I think I'm approaching this from a REST perspective, but I find this user is a separate entity as the users we use elsewhere in general. But honestly, this probably fits the graphql paradigm more and we can just not specify emailVerified in the query.

mslwang: I'll add it to the UserDTO since the frontend deals with the data differently than the backend anyway.

frontend/src/types/AuthTypes.ts
mslwang: > Let's keep it the same if possible, if you don't mind reverting everything back? If you feel strongly about it we can make a separate ticket and PR for that change.

Good to just have this happen in a separate PR.

backend/types.ts
```diff
@@ -38,11 +38,7 @@ export type UpdateUserDTO = Omit<UserDTO, "id">;
 
 export type RegisterUserDTO = Omit<CreateUserDTO, "role">;
 
-export type UserWithVerificationStatusDTO = UserDTO & {
```
mslwang: Wondering if this would still be useful at this point

mslwang:
> Was there a conclusion on this? Maybe becomes a nitpicky product question, but I think a better user experience would be limiting functionality for accounts with unverified emails, without a hard block on them looking around. Admittedly that might be a lot more dev work, but I don't think the whole "nothing here" screen is an especially welcoming experience for a new user that could just be looking to poke around

I agree that it'd be better if there wasn't a hard block on entry here, with the detail that we should still make it clear that they have to verify their email (the latter is important because I remember having MongoDB onboarding issues because I wasn't aware that I had to verify my email, given that some websites don't even need email verification). But, I brought up building the user flow in this direction with @jfdoming and we decided to put it off to prioritize deal-breaking features. 

Regarding how we could alternatively do this, I don't think I've roped @draymond63 into this, but just to spitball, I think letting them snoop around with clear knowledge of their unverified status is the way to go. I was looking around for inspiration, and I saw that Reddit has this as their user flow. After signing up and going through the recommendation onboarding, you are left at the feed with a little popup next to the user. 
image
After you click "Got it", a modal shows up that an email has been sent that can be cleared out.
image

I think we could have a persistent little header that always shows up at the top that says "verify email" that I feel like I've seen before.

jfdoming:
> > Was there a conclusion on this? Maybe ...

This is a great idea! I'll definitely create a ticket for this, but as you mentioned, probably best to de-prioritize this work for now. One additional thought I had is we should decide what parts of the app to specifically restrict to unverified users. For example, should unverified learners have access to actually go through courses? Should unverified admins be able to modify/delete courses or users? It might make sense to drop unverified users into a "read-only" view regardless of their role. Definitely we can discuss this in the future if this ticket gets a chance to be worked on, though.

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -142,49 +143,55 @@ const refresh = async (refreshFunction: RefreshFunction): Promise<boolean> => {
...
+  setAuthUser: Dispatch<SetStateAction<AuthUser>>,
+): void => {
+  setAuthUser((prevAuthUser: AuthUser) => {
+    const newAuthUser = {
+      ...prevAuthUser,
+      emailVerified,
+    } as AuthUser;
+    localStorage.setItem(AUTHENTICATED_USER_KEY, JSON.stringify(newAuthUser));
+    return newAuthUser;
+  });
```
jfdoming, line 187: I don't think it makes sense to have this function located in this file. Why not include this logic directly at the call site? Or if you feel it's necessary to extract it into a function, could you put the function directly in the file where its used? The APIClient files are really just for functionality which makes calls to the backend.

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -4,32 +4,37 @@ import { useQuery } from "@apollo/client";
...
-        }
-      });
+    if (authUserEmail) {
+      authAPIClient
+        .getEmailVerified(authUserEmail, refetch)
+        .then((emailVerified) => {
+          if (
+            emailVerified !== null &&
+            authUserEmailVerified !== emailVerified
+          ) {
```
jfdoming, line 29: Are these checks actually necessary? The dispatcher call should be very efficient, so unless doing so breaks the functionality, I would remove this check for code cleanliness purposes. (Incidentally, doing so will also allow you to remove the `authUserEmailVerified` constant which would make the code even shorter.)

frontend/src/contexts/AuthContext.ts
```diff
@@ -1,15 +1,17 @@
-import { createContext } from "react";
+import { createContext, Dispatch, SetStateAction } from "react";
 import { AuthUser } from "../types/AuthTypes";
 
 type AuthContextType = {
   authUser: AuthUser;
-  setAuthUser: (_authUser: AuthUser) => void;
+  setAuthUser: Dispatch<SetStateAction<AuthUser>>;
```
jfdoming, line 6: Nice!

frontend/src/types/AuthTypes.ts
jfdoming: Would you be able to revert these name changes then, @mslwang ?

frontend/src/components/pages/VerifyEmail.tsx
mslwang: Yeah, I included something about it. Basically setting it every time made it so that the useEffect was called again which would keep running it until we got rate limited by the backend.

frontend/src/types/AuthTypes.ts
mslwang: sounds good

frontend/src/components/pages/VerifyEmail.tsx
mslwang: image
example

jfdoming (approved):
Approving, but one last thought: it would be a nice UX if the page redirected the user automatically to their homepage once they are verified. How would you feel about adding a `useInterval` call to refetch the user periodically (say once every 15-30 seconds)? If you're on board with that, feel free to make a separate ticket to track it (you can use the implementation [here](https://codesandbox.io/s/l240mp2pm7)).

PR #166, opened 2022-06-04
Use Vault for our team secrets

jfdoming:
## Implementation Description
We now have a Vault KV-store set up for our team! This PR updates the dev environment to fetch secrets from our Vault, so that we don't have to put secrets on our public Notion.

- Update post-merge hook to be RHS-specific
- Update setup script name to be more clear
- Add local environment file so people can use their own MongoDB

## Screenshots
N/A

- 

## Testing Procedure

## Things to Note

## Checklist


e-wai:
Could you add a quick "how-to" note for setting up the repo? (The README is not super helpful about this part)

jfdoming:
> Could you add a quick "how-to" note for setting up the repo? (The README is not super helpful about this part)

Updated, thanks for pointing this out! I had updated the Notion docs but it completely slipped my mind that the README was also a source of documentation. 🤦‍♂️

mslwang:
I don't have the full context, by why was the setup file replaced? It seems like it's still the same in starter code.

jfdoming:
> I don't have the full context, by why was the setup file replaced? It seems like it's still the same in starter code.

It should be a rename, not a replacement, but the main change is that we no longer need to update the hook on every run, and the script will actually execute the hook now.

mslwang:
> > I don't have the full context, by why was the setup file replaced? It seems like it's still the same in starter code.
> 
> It should be a rename, not a replacement, but the main change is that we no longer need to update the hook on every run, and the script will actually execute the hook now.

Is that reflected in the general documentation with that in mind? Or is this specific to our project?

jfdoming:
> > > I don't have the full context, by why was the setup file replaced? It seems like it's still the same in starter code.
> > 
> > 
> > It should be a rename, not a replacement, but the main change is that we no longer need to update the hook on every run, and the script will actually execute the hook now.
> 
> Is that reflected in the general documentation with that in mind? Or is this specific to our project?

The general documentation doesn't mention the details of this setup script. It's present but not documented in the starter code repository. This change is specific to our project but I might try to upstream these changes if people like it.

jfdoming:
Closing as not necessary.

PR #169, opened 2022-06-09
[RHS-136]: Implement Editing User Role

a4bello:
## Implementation Description
Admin users should be able to change the role of users between `User`, `Staff`, and `Admin`. From the Manage Users , a user can be searched up and the admin can change their role. The Edit Role button should open a modal that allows this functionality to be achieved. The modal component is based off the `frontend/src/components/common/Modal.tsx` component. A `SelectInput` component was also created to create a reusable component for select drop-down lists. This component is used in the modal to choose the role the admin wants the user to become. 

A userMutation graphql file was also created to include the userUpdate Mutation necessary to update the user's role in the database. 

## Screenshots
image
image
image

Reviewers should focus on ensuring the code is clean and written to the highest quality. Reviewers should also ensure the following functionalities is achieved:
- Only Admin Users can edit the role of other users (including other admins).
- If a user's role is successfully changed, there should be a change in the user card displayed for the user. 
- The change in user role is evident on the database. 

## Testing Procedure

## Things to Note

## Checklist


frontend/src/components/admin/UserCard.tsx
```diff
@@ -27,12 +42,42 @@ const DefaultUserIcon = ({
...
+  id,
   firstName,
   lastName,
   role,
   email,
   town,
 }: UserCardProps): React.ReactElement => {
+  const userCardOptions = ["User", "Admin", "Staff"];
+  const [selectValue, setSelectValue] = useState(userCardOptions[0]);
+  const [currRole, setCurrRole] = useState(role);
```
jfdoming, line 54: This solution almost works, but unfortunately, if you switch users (email search) and switch back, the role does not seem to update properly. It might be simpler instead to just make use of `refetchQueries`.

frontend/src/components/admin/UserCard.tsx
```diff
@@ -55,15 +100,34 @@ const UserCard = ({
...
+          onConfirm={() => updateUserRender(selectValue)}
+          onCancel={() => {
+            resetState();
+            onClose();
+          }}
+        >
+          <SelectInput
+            onChange={(currValue) => {
+              setSelectValue(currValue);
+            }}
```
jfdoming, line 118:
```suggestion
            onChange={setSelectValue}
```

frontend/src/APIClients/mutations/UserMutations.ts
```diff
@@ -0,0 +1,16 @@
...
+const UPDATE_USER = gql`
+  mutation UpdateUser($id: ID!, $user: UpdateUserDTO!) {
+    updateUser(id: $id, user: $user) {
+      id
+      firstName
+      lastName
+      email
+      town
+      role
+    }
```
jfdoming, line 12: nit: this is a lot of fields to pass to the backend. It might be better to make a specific `UpdateUserRole` mutation for this purpose.

jfdoming (changes requested):
Great job on this, especially with the super fast turnaround! I left a few comments to think about, but overall, excellent work.

frontend/src/APIClients/mutations/UserMutations.ts
mslwang: Yeah, could be good to scope it here from the frontend, but I find it still ends up being somewhat redundant for any other updates.

frontend/src/components/admin/UserCard.tsx
```diff
@@ -48,30 +52,32 @@ const UserCard = ({
...
+    setSelectValue(role);
+  }, [role]);
+
+  const [UpdateUserRole] = useMutation(UPDATE_USER_ROLE);
 
-  const resetState = () => {
-    setSelectValue(userCardOptions[0]);
+  const resetState = (userRole: Role) => {
+    setSelectValue(userRole);
   };
```
jfdoming, line 68:
```suggestion
  const resetState = setSelectValue;
```

Or just use `setSelectValue` directly.

frontend/src/components/common/SelectInput.tsx
```diff
@@ -15,8 +15,8 @@ const SelectInput = ({
   return (
     <Select value={value} onChange={(e) => onChange(e.target.value)}>
       {optionsMap.map((options) => (
-        <option key={options} value={options}>
-          {options}
+        <option key={options.value} value={options.value}>
+          {options.label}
```
jfdoming, line 19: Nice!

frontend/src/components/pages/ManageUsersPage.tsx
```diff
@@ -67,6 +65,7 @@ const ManageUsersPage = (): React.ReactElement => {
               role={spotlightedUser.role}
               email={spotlightedUser.email}
               town={spotlightedUser.town}
+              onUserSelect={(user: UserResponse) => setSpotlightedUser(user)}
```
jfdoming, line 68: 👌

frontend/src/constants/UserConstants.ts
```diff
@@ -0,0 +1,3 @@
+const USER_ROLES = ["User", "Admin", "Staff"];
```
jfdoming, line 1:
```suggestion
const USER_ROLES = ["User", "Admin", "Staff"] as const;
```

If it works, the `as const` lets Typescript know that we want to restrict to these specific strings.

jfdoming (changes requested):
A few minor nits, but nice refactors!

jfdoming (approved):
Nice job!

PR #170, opened 2022-06-09
[RHS-168] Rename Authentication User Type in Frontend

mslwang:
## Implementation Description
Renamed all instances of `AuthenticationUser` to `AuthUser`

- Whether the code still works the same as the status quo

## Testing Procedure
   - Everything you'd normally do

## Checklist

### DIFF
--- a/frontend/src/APIClients/AuthAPIClient.ts
+++ b/frontend/src/APIClients/AuthAPIClient.ts
@@ -99,11 +99,11 @@ type LogoutFunction = (
 >;
 
 const logout = async (
-  authenticatedUserId: string,
+  authUserId: string,
   logoutFunction: LogoutFunction,
 ): Promise<boolean> => {
   const result = await logoutFunction({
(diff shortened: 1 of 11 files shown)


jfdoming:
My understanding of [this comment](https://github.com/uwblueprint/rowan-house/pull/160#discussion_r890690828) was that this change was unnecessary, since a user can be "authenticated" but not "verified" at the same time; hence `AuthenticatedUser` still applies even if a user is unverified. Was that not your interpretation?

mslwang:
> My understanding of [this comment](https://github.com/uwblueprint/rowan-house/pull/160#discussion_r890690828) was that this change was unnecessary, since a user can be "authenticated" but not "verified" at the same time; hence `AuthenticatedUser` still applies even if a user is unverified. Was that not your interpretation?

Hmmm, yeah, I think I understand this perspective with being logged in meaning that you are authenticated to maneuver around routes. In that case, I'm just going to remove this.

PR #174, opened 2022-06-12
[RHS-173]: Enable debugging

jfdoming:
## Implementation Description
A feature for the development environment requested by devs was to be able to connect to the frontend/backend via a debugger. This PR adds some configuration to the repository to support this.
- [RHS-173]: Enable debug mode for node while developing
- [RHS-173]: Add launch configurations for devs using VS Code

## Screenshots
N/A

- How do you feel about permanently running in `--inspect` mode for all local development? Is it worth it in order to prevent having to commit a `settings.json` with extra command configuration?

## Testing Procedure
Feel free to check this out locally and try it yourself. The normal development flow is unchanged (run `docker-compose up --build`), but you should be able to run additional debug configurations from the tab on the left-hand side of VS Code.

## Things to Note
N/A

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,8 @@
 **/*.log
 **/firebaseServiceAccount.json
 **/.DS_Store
-.vscode
+.vscode/**/*
+!.vscode/launch.json
 **/*.cache
 **/*.egg-info
 .history/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,22 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug: Frontend",
+            "type": "pwa-chrome",
+            "request": "launch",
+            "url": "http://localhost:3000",
+            "webRoot": "${workspaceFolder}/frontend",
+            "sourceMapPathOverrides": {
+                "/app/*": "${webRoot}/*"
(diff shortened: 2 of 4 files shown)


mslwang (approved):
lgtm

mslwang:
Could be good to add some instructions to the README just in case someone hasn't used a vscode debugger before, but it's pretty straightforward overall, so we could at least state its existence in the README.

jfdoming:
> Could be good to add some instructions to the README just in case someone hasn't used a vscode debugger before, but it's pretty straightforward overall, so we could at least state its existence in the README.

Yep, I updated the Notion workspace documentation in the Onboarding section with this information, since it didn't seem to fit in the README based on the current content. I think it would be better to direct people to the Notion workspace if possible, because that way we can keep all our docs in one place.

mslwang:
> > Could be good to add some instructions to the README just in case someone hasn't used a vscode debugger before, but it's pretty straightforward overall, so we could at least state its existence in the README.
> 
> Yep, I updated the Notion workspace documentation in the Onboarding section with this information, since it didn't seem to fit in the README based on the current content. I think it would be better to direct people to the Notion workspace if possible, because that way we can keep all our docs in one place.

Nice, just checked the notion, looks good!

PR #177, opened 2022-06-15
Content block modals

molly-yu:
## Implementation Description
Implements #112. 
Created an edit text modal that allows users to update text content blocks using the `update-lesson` action. The content block should update properly and allow users to save their changes afterwards.

## Screenshots
image

- 

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/components/admin/EditTextModal.tsx
+++ b/frontend/src/components/admin/EditTextModal.tsx
@@ -27,16 +27,24 @@ const EditTextModal = ({
   const { state, dispatch } = context;
 
   const onSave = () => {
-    const newBlock = block;
-    newBlock.content.text = text;
+    const newBlock = {
+      ...block,
+      content: {
+        ...block.content,
+        text,
+      }
+    };
+    // newBlock.content.text = text;
     dispatch({
       type: "update-block",
       value: {
         index,
-        block,
+        block: newBlock,
       },
     });
-  }
+    // state.
+    onClose();
+  };
 
   return (
     // eslint-disable-next-line react/jsx-props-no-spreading
--- a/frontend/src/components/admin/EditTextModal.tsx
+++ b/frontend/src/components/admin/EditTextModal.tsx
@@ -1,0 +1,59 @@
+import { Flex, VStack, Image } from "@chakra-ui/react";
+import React, { useContext, useState } from "react";
+
+import { TextInput } from "../common/TextInput";
+import { Modal } from "../common/Modal";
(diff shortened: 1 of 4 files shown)


frontend/src/components/module-editor/EditableContentBlock.tsx
```diff
@@ -54,12 +62,22 @@ const EditableContentBlock = ({
             {SelectContentBlock(block)}
             <EditContentOptionsMenu
               isVisible={isHovered}
-              onEditClick={() => {}}
+              onEditClick={() => {
+                onOpen();
+              }}
```
jfdoming, line 67: Can you pass `onOpen` directly here? i.e.,
```suggestion
              onEditClick={onOpen}
```

frontend/src/components/admin/EditTextModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <Modal
+      size="xl"
+      header="Edit Text Component"
+      onConfirm={onSave}
+      onCancel={onClose}
+      isOpen={isOpen}
+    >
+      <Flex>
+        <VStack flex="1">
+          <TextArea defaultValue={text} onChange={setText} isRequired />
```
jfdoming, line 68: Unfortunately, just setting `isRequired` doesn't actually validate that the field contains text. Would you be able to add some validation in `onSave`? You can set the `errorMessage` prop if there's an error.

Also, it would be good for accessibility to add a label to this field (pass a `label` prop).

frontend/src/components/module-editor/LessonViewer.tsx
```diff
@@ -20,7 +20,12 @@ const LessonViewer = (): React.ReactElement => {
         {(provided) => (
           <Flex ref={provided.innerRef} direction="column" flex="1">
             {lesson?.content.map((block, index) => (
-              <EditableContentBlock block={block} key={index} index={index} />
+              <EditableContentBlock
+                block={block}
+                key={index}
+                index={index}
+                context={context}
```
jfdoming, line 27: Instead of passing `context` all the way down, it would be better to add a `useContext` hook call inside `EditTextModal`.

jfdoming (changes requested):
Nice work! Left a few comments.

jfdoming (commented):
Just one small comment!

jfdoming:
In the interest of merging this to unblock future work, I'm resolving my own PR comment and merging this.

jfdoming (approved):

PR #178, opened 2022-06-16
[RHS-139] Add Verification Email Resend flow

mslwang:
## Implementation Description
Added ability to resend verification email in the event that old verification email has expired.

## Screenshots
Default verify email page
image

Successfully re-sent verification email
image

Failed to send verification email
image

- Correctness of code

## Testing Procedure
- Try going through the workflow and seeing if you can make it work successfully or make it error. See if anything can be broken.

## Things to Note
- We should look to add explicit rate limiting

## Checklist

### DIFF
--- a/backend/graphql/resolvers/authResolvers.ts
+++ b/backend/graphql/resolvers/authResolvers.ts
@@ -83,8 +83,8 @@ const authResolvers = {
         return true;
       } catch (error: unknown) {
         const errorMessage = [
-          `Failed to resolve email verification link mutation for user with email ${email}`,
-          "Reason =",
+          `Failed to resolve email verification link mutation for user with email ${email}, `,
+          "reason =",
           getErrorMessage(error),
         ];
         Logger.error(errorMessage.join(" "));
--- a/backend/graphql/resolvers/authResolvers.ts
+++ b/backend/graphql/resolvers/authResolvers.ts
@@ -8,6 +8,10 @@ import IAuthService from "../../services/interfaces/authService";
 import IEmailService from "../../services/interfaces/emailService";
 import IUserService from "../../services/interfaces/userService";
 import { AuthDTO, RegisterUserDTO, Role } from "../../types";
+import { getErrorMessage } from "../../utilities/errorUtils";
+import logger from "../../utilities/logger";
+
+const Logger = logger(__filename);
 
 const userService: IUserService = new UserService();
 const emailService: IEmailService = new EmailService(nodemailerConfig);
@@ -69,6 +73,23 @@ const authResolvers = {
     ): Promise<boolean> => {
       await authService.resetPassword(email);
       return true;
+    },
+    sendEmailVerificationLink: async (
+      _parent: undefined,
+      { email }: { email: string },
+    ): Promise<boolean> => {
+      try {
+        await authService.sendEmailVerificationLink(email);
+        return true;
+      } catch (error: unknown) {
+        const errorMessage = [
+          `Failed to resolve email verification link mutation for user with email ${email}`,
+          "Reason =",
+          getErrorMessage(error),
+        ];
+        Logger.error(errorMessage.join(" "));
+        return false;
+      }
     },
   },
 };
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -1,21 +1,78 @@
...
 import authAPIClient from "../../APIClients/AuthAPIClient";
+import { SEND_EMAIL_VERIFICATION_LINK } from "../../APIClients/mutations/AuthMutations";
+
+interface VerifyEmailProps {
+  hasError: boolean;
+  hasBeenPressed: boolean;
+  setHasBeenPressed: Dispatch<SetStateAction<boolean>>;
+}
+
+const VerifyEmailContent = ({
```
mslwang, line 24: Wasn't sure where I should put this component so I left it in the same file, feel free to let me know where I should put this.

backend/graphql/resolvers/authResolvers.ts
```diff
@@ -70,6 +74,23 @@ const authResolvers = {
...
+    sendEmailVerificationLink: async (
+      _parent: undefined,
+      { email }: { email: string },
+    ): Promise<boolean> => {
+      try {
+        await authService.sendEmailVerificationLink(email);
+        return true;
+      } catch (error: unknown) {
+        const errorMessage = [
+          `Failed to resolve email verification link mutation for user with email ${email}`,
```
jfdoming, line 86: If these strings are being joined with a " ", it might make sense to add a trailing comma to this line

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -1,21 +1,78 @@
...
+      </Button>
+    </>
+  );
+};
 
 const VerifyEmail = (): React.ReactElement => {
   const { authenticatedUser, setAuthenticatedUser } = useContext(AuthContext);
+
+  const [hasBeenPressed, setHasBeenPressed] = useState(false);
+  const [hasError, setHasError] = useState(false);
```
jfdoming, line 64: Rather than having several different flags, it could make sense to just keep one "state" variable since these states are all mutually exclusive. It might make the logic clearer (i.e., we're either in the "default" state, the "sent" state or the "error" state). Just an optional suggestion though, I don't mind this solution as is since there are only two flags.

frontend/src/components/pages/VerifyEmail.tsx
```diff
@@ -26,18 +83,44 @@ const VerifyEmail = (): React.ReactElement => {
...
+          });
+          if (result) {
+            const success = result.data?.sendEmailVerificationLink ?? null;
+            if (success) {
+              // TODO: Add rate limiting for button presses
+              setHasBeenPressed(true);
+              return;
+            }
+            setHasError(true);
+          }
```
jfdoming, line 101: Currently, if there's no `result`, this code won't set an error or success. Is that intended? If not, you can probably combine these checks into one simplified check to reduce nesting.

frontend/src/components/pages/VerifyEmail.tsx
jfdoming: Seems fine to me to keep it here, this is a short file anyways.

jfdoming (commented):
Nice job! Left some comments, let me know your thoughts.

frontend/src/components/pages/VerifyEmail.tsx
mslwang: Yeah, I thought this too, definitely better logically to change it, but I didn't think about it much more.

frontend/src/components/pages/VerifyEmail.tsx
mslwang: Was a mental carryover from how I did it in AuthAPIClient, will fix.

frontend/src/components/pages/VerifyEmail.tsx
mslwang: Yeah, more like I know by principle, some people like having one component per file.

jfdoming (approved):
Nice!

PR #180, opened 2022-06-21
Image upload for modules

y2kwak:
## Implementation Description
Adding image upload functionality for module preview images!
There are two image related fields on a `module`: `previewImage` and `fileName`. The `previewImage` is a signedUrl that can be used to display the image, and the `fileName` is the file path and main indicator for the image uploaded.

Note that the logic for uploading a module image is separate from the other course mutations, such as `createCourse` and `updateCourse`.

## Screenshots
Screen Shot 2022-07-05 at 00 18 02
Screen Shot 2022-07-05 at 00 18 09

## Testing Procedure

## Things to Note

## Checklist

## To-do:

- [x] implement upload image logic (be)
- note: the logic is separate from create/update course mutation
- [x] show uploaded image in modal (after user selects an image, but before they click 'confirm')
- [x] show uploaded image in modal preview
- [x] verify uploaded image is good quality


jfdoming (commented):
Broadly speaking, wondering about this: we currently have a separate endpoint to fetch the image which is only called *after* the page loads. Is there any reason not to do this logic *when* the page loads instead? i.e., provide the image with the COURSES query (since whenever we fetch ALL courses, we also want the module's image). To make the module editor work, we can update GET_COURSE to take in an optional list of module indices to fetch images for.

With the above logic, the UX is a bit better, since the user only has to wait for _one_ loading spinner rather than some additional loading afterwards.

Also, by using the above logic, we can remove the need for two separate `fileName` and `previewImage` fields; instead, we can use the `previewImage` field in the DB for the filename, and process it when returned to the frontend to contain the real URL (or null if not fetched).

Let me know your thoughts on this!

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -83,6 +109,12 @@ const courseResolvers = {
 
       return courseService.getCourses(attributes);
     },
+    moduleImage: async (
```
jfdoming, line 112: I don't think this endpoint is used.

frontend/src/components/common/EditModuleModal.tsx
```diff
@@ -54,17 +66,51 @@ const EditModuleModal = ({
...
-    setInvalid(false);
-    onClose();
+
+    if (previewImage) {
+      const newModule = {
+        ...module,
+        title,
+        description,
+        published: isPublished,
+        previewImage,
```
jfdoming, line 76: Instead of setting `previewImage` to the actual preview URL here (which expires after 60 minutes), I'm thinking it would be good to set it to the file path, and hydrate `filePath -> previewImage` later during `GET_COURSE` or `COURSES` queries. That's because currently the image URLs break after that hour passes. Thoughts?

frontend/src/components/common/EditModuleModal.tsx
```diff
@@ -54,17 +66,51 @@ const EditModuleModal = ({
...
-      variables: { id, course },
-      refetchQueries: [
-        { query: COURSES },
-        { query: GET_COURSE, variables: { id } },
-      ],
-    });
-    setInvalid(false);
-    onClose();
+
+    if (previewImage) {
```
jfdoming, line 70: Probably best to skip this check, I don't think images are required for modules.

jfdoming (commented):
(not sure if this was ready for review, but I left some comments to hopefully help out a bit) Awesome, thanks for figuring this out! Wondering about a few things, also not sure if you ran into this or if you already knew about this but I encountered an issue on your branch where the images weren't showing up, you probably need to rename `preview_image` in the model schema to `previewImage`.

frontend/src/components/common/EditModuleModal.tsx
```diff
@@ -78,7 +124,46 @@ const EditModuleModal = ({
...
+          <Box
+            borderRadius=".5rem"
+            onMouseEnter={() => setIsHover(true)}
+            onMouseLeave={() => setIsHover(false)}
+            _hover={{
+              background: "black",
+              opacity: 0.9,
+            }}
+            onClick={openFileBrowser}
+          >
```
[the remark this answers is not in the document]
smilesthesad: agreed

backend/services/interfaces/IImageUploadService.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface UploadedImage {
+  image: string;
+  path: string;
+}
+
+interface IImageUploadService {
+  /**
+   * Upload an image to Firebase.
+   * @param file the file to upload
+   * @returns a link and filename for the uploaded image
```
Tayef-Shah, line 12: nit: based on the UploadedImage interface, it might be more accurate to say
```suggestion
   * @returns a link and file path for the uploaded image
```

frontend/src/components/common/EditModuleModal.tsx
```diff
@@ -1,4 +1,4 @@
-import { Flex, VStack, Image } from "@chakra-ui/react";
+import { Box, Flex, VStack, Image, Text } from "@chakra-ui/react";
```
Tayef-Shah, line 1: nit: linter

frontend/src/components/common/EditModuleModal.tsx
```diff
@@ -64,9 +86,35 @@ const EditModuleModal = ({
...
+    if (e.target.files) {
+      const fileSize = e.target.files[0].size / 1024 / 1024;
+      if (fileSize > 5) {
+        // eslint-disable-next-line no-alert
+        window.alert("Your file exceeds 5MB. Upload a smaller file.");
+      } else {
+        const imageUploadResult = await uploadImage({
+          variables: { file: e.target.files[0] },
+        });
+        console.log(imageUploadResult);
```
Tayef-Shah, line 103: might wanna use the logger here

frontend/src/components/common/EditModuleModal.tsx
```diff
@@ -78,7 +124,46 @@ const EditModuleModal = ({
     >
       <Flex>
         <VStack flex="1" pr="1rem">
-          <Image src={DEFAULT_IMAGE} borderRadius=".5rem" />
+          <Box
```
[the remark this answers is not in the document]
Tayef-Shah: yeah agreed. There's already a lot going on in this component, so moving the image uploading to a different component would be clean

backend/graphql/resolvers/courseResolvers.ts
```diff
@@ -18,6 +22,12 @@ import { Role } from "../../types";
 import { CourseVisibilityAttributes } from "../../models/course.model";
 import { assertNever } from "../../utilities/errorUtils";
 
+const defaultBucket = process.env.FIREBASE_STORAGE_DEFAULT_BUCKET || "";
+const fileStorageService = new FileStorageService(defaultBucket);
```
Tayef-Shah, line 26: what would happen here if there's no env variable for the default bucket? Would you still be able to create a FileStorageService with a bucket of ""?

backend/services/implementations/imageUploadService.ts
```diff
@@ -0,0 +1,86 @@
...
+      return await this.storeImage(filePath, fileContentType);
+    } finally {
+      if (filePath) {
+        fs.unlinkSync(filePath);
+      }
+    }
+  }
+
+  async download(fileName: string): Promise<string> {
+    return this.storageService.getFile(fileName, 60);
```
Tayef-Shah, line 61: the 60 is a bit of a magic number at the moment - might want to define it as a constant or a variable so people will know what it represents

Tayef-Shah (dismissed):
Seems fine overall. I'll approve it since we should try to test it soon and I don't think anyone else is going to review.

backend/graphql/resolvers/courseResolvers.ts
jfdoming: You're right that this wouldn't work in that case. However, since we're in control of the environment, I don't think this case would ever occur.

Tayef-Shah: I see. I guess no need for the || "" in that case, but no harm in leaving it in.

PR #182, opened 2022-06-21
Set up front-end structure to support more content types

jfdoming:
## Implementation Description
Major refactors to generally support the assorted content types. Overview:
- Content block state is now strongly typed
- There are now dedicated locations for content block renderers and editing modals

## Screenshots
No visual changes (hopefully)

- Do the refactors make sense? Will they make it easier to add new content types?

## Testing Procedure
Use the editor, verify nothing is broken.

## Things to Note
N/A

## Checklist

### DIFF
--- a/frontend/src/components/common/content/ContentBlockRenderer.ts
+++ b/frontend/src/components/common/content/ContentBlockRenderer.ts
@@ -1,6 +1,6 @@
 import React from "react";
-import { ContentType } from "../APIClients/types/LessonClientTypes";
-import { ContentBlockState } from "../types/ModuleEditorTypes";
+import { ContentType } from "../../../APIClients/types/LessonClientTypes";
+import { ContentBlockState } from "../../../types/ContentBlockTypes";
 
 // Many thanks to https://github.com/Microsoft/TypeScript/issues/30581#issuecomment-1080979994.
 
--- a/frontend/src/components/common/content/ContentBlockRenderer.ts
+++ b/frontend/src/components/common/content/ContentBlockRenderer.ts
@@ -78,7 +78,7 @@ interface ConfigEntry<T extends ContentType> {
     | typeof EmptyConfigComponent;
 }
 
-export const EmptyConfigComponent = () => null;
+export const EmptyConfigComponent = (): null => null;
 export const EmptyConfigEntry = {
   renderBlock: EmptyConfigComponent,
   renderEditModal: EmptyConfigComponent,
--- a/frontend/src/components/common/content/ContentBlockRenderer.ts
+++ b/frontend/src/components/common/content/ContentBlockRenderer.ts
@@ -1,13 +1,21 @@
 import React from "react";
 import { ContentType } from "../../../APIClients/types/LessonClientTypes";
-import { ContentBlockState } from "../../../types/ContentBlockTypes";
+import {
+  ContentBlockProps,
+  ContentBlockState,
+  ContentBlockTypeToState,
+} from "../../../types/ContentBlockTypes";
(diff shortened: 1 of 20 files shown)


frontend/src/components/common/content/ContentBlockRenderer.ts
```diff
@@ -0,0 +1,64 @@
...
+  ): block is ValueOf<ArgMap<DispatchableT>> {
+    return block.type.clientType in this.dispatchTable;
+  }
+
+  public render(props: PropType): React.ReactNode {
+    const { block, ...rest } = props;
+
+    if (this.canDispatch(block)) {
+      return this.callWithArgMap({ block, ...rest });
+    }
```
draymond63, line 56: Can this not be written more simply? I get that this was copied from the github issue, but it could be improved :)
```javascript
const { block } = props;

if (this.canDispatch(block)) {
   return this.callWithArgMap(props);
}
```

frontend/src/components/common/content/ContentBlockRenderer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+// Many thanks to https://github.com/Microsoft/TypeScript/issues/30581#issuecomment-1080979994.
+
+type ValueOf<T> = T[keyof T];
+
+type ArgMap<D extends ContentType> = {
+  [K in D]: Extract<ContentBlockState, { ["type"]: { ["clientType"]: K } }>;
+};
+
+class ContentBlockRenderer<
```
draymond63, line 13: It might be good to have a comment talking about why this is necessary so it's less cryptic (e.g. so we can have a table instead of a switch case or "Implements a table that takes in a component block type and renders a coressponding React component")

frontend/src/components/module-editor/content/EditableContentBlock.tsx
```diff
@@ -2,35 +2,52 @@ import { Box, Divider, Flex, useDisclosure, VStack } from "@chakra-ui/react";
...
+>({
+  column: Empty,
+  heading: Empty,
+  text: EditTextModal,
+  link: Empty,
+  button: Empty,
+  image: EditImageModal,
+  video: Empty,
+  audio: Empty,
+});
```
draymond63, line 44: It could be nice to define one large table with: 
1. Content type (e.g. text)
2. Content block component (e.g. TextBlock)
3. Edit modal (e.g. EditTextModal)

Then we could automagically split this larger table up into the two ContentBlockRenderer objects you have here. That way people can create a new component and put everything into one "row" of the table as opposed to editing multiple things in the content block. Is it worth the possible additional complexity? Idk, but worth a thought.

draymond63 (approved):
I think it looks great! A few comments here and there, but the logic makes sense to me.

frontend/src/components/common/content/ContentBlockRenderer.ts
jfdoming: Unfortunately, no, actually! Or at least, I wasn't able to figure out how to make it work. If you implement the solution as given above, VSCode complains:
image

I think it's because with the current solution, Typescript knows that the type of `block` was narrowed, but it's not smart enough to realize that the `block` which we extracted is the same `block` as is contained in `props`. I could just cast it and avoid the extra verbosity, but I wanted to avoid casts here if possible.

draymond63: Dang, seems like sketchy typescript stuff to me. Good to know - I'm happy with it then :)

draymond63 (approved):
I like it! Much better than the switch case implementation.

PR #183, opened 2022-06-28
[RHS-117]: Fix main-page logout button

jfdoming:
## Implementation Description
- [RHS-117]: Extract logout button to new component
- Minor style fix for the logout button
- [RHS-117]: Add logout page

## Screenshots
image
image

- 

## Testing Procedure
Try to log out, verify it's intuitive and works.

## Things to Note
N/A

## Checklist

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -16,6 +16,7 @@ import PrivateRoute from "./components/auth/PrivateRoute";
 import CreatePage from "./components/pages/CreatePage";
 import Default from "./components/pages/Default";
 import DisplayPage from "./components/pages/DisplayPage";
+import Logout from "./components/pages/Logout";
 import NotFound from "./components/pages/NotFound";
 import UpdatePage from "./components/pages/UpdatePage";
 import ModuleEditor from "./components/pages/ModuleEditor";
@@ -121,6 +122,11 @@ const App = (): React.ReactElement => {
                     exact
                     path={`${Routes.VERIFY_EMAIL_PAGE}`}
                     component={VerifyEmail}
+                  />
+                  <Route
+                    exact
+                    path={`${Routes.LOGOUT_PAGE}`}
+                    component={Logout}
                   />
                   <Route exact path="*" component={NotFound} />
                 </Switch>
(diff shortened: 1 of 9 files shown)


frontend/src/App.tsx
```diff
@@ -122,6 +123,11 @@ const App = (): React.ReactElement => {
                     path={`${Routes.VERIFY_EMAIL_PAGE}`}
                     component={VerifyEmail}
                   />
+                  <Route
+                    exact
+                    path={`${Routes.LOGOUT_PAGE}`}
+                    component={Logout}
+                  />
```
draymond63, line 130: What's the point of this page? Just for development purposes?

frontend/src/hooks/useLogout.ts
```diff
@@ -1,36 +1,34 @@
...
 
+  const history = useHistory();
+
   const [logout] = useMutation<{ logout: null }>(LOGOUT);
 
-  const onLogOutClick = async () => {
+  const tryLogout = async () => {
     try {
       await authAPIClient.logout(String(authenticatedUser?.id), logout);
+      await tryLogout();
```
draymond63, line 23: So this will call itself until it fails, and only then will it jump into the catch handler and exit the function? Like, we are intentionally failing/expecting an error, because that means we aren't authenticated anymore. This seems like an interesting/counterintuitive way to go about it. What happens if the authAPIClient.logout fails an we continue to be logged in? Will we be stuck in an infinite recursive loop?

draymond63 (commented):
A few questions, mainly about the new hook and the new logout page. Thanks Julian!

draymond63:
It also seems like the preview link is broken. Do you know why that is?

frontend/src/App.tsx
jfdoming: Yeah, and also to have a page to direct people to in case something gets screwed up with their auth. So if someone is getting page-breaking auth errors, we can tell them "visit /logout" and try again. (Certain auth errors cause the entire page, including the logout button, to disappear.)

frontend/src/hooks/useLogout.ts
jfdoming: Yeah, exactly; that's how the existing logic works at least. I considered limiting it to a certain number of tries to log out, but not sure if that's worth it. It should usually fail in this case. Additionally I could add an additional check on the return value of `authAPIClient.logout`.

jfdoming:
> It also seems like the preview link is broken. Do you know why that is?

That I could not tell you... let me try making a minimal change and try again.

Issue #136, opened 2022-05-17
Dashboard: Editing user role


draymond63:
## Summary
Admins have the ability change the role of users between `User`, `Staff`, and `Admin`. This ticket outlines the UI and API interaction required to get this working!

![image.png](https://images.zenhubusercontent.com/6053e4ca8db077e41d15aa4a/aaad9773-fed5-478f-a7c2-360edd01e055)
![image.png](https://images.zenhubusercontent.com/6053e4ca8db077e41d15aa4a/823a2024-8a35-4b11-8a45-10a5d95aaa14)

## Goals
- [x] Clicking "Edit Role" opens a modal 
- [x] Modal allows you to select a role from a dropdown menu
- [x] Confirming the choice calls the `updateUser` mutation, changing the role

## Implementation
Modal should be made based on the `frontend/src/components/common/Modal.tsx` component. You can use the [Select](https://chakra-ui.com/docs/components/form/select) component from Chakra-UI for the role dropdown.

PR #184, opened 2022-07-02
[RHS-179] : Prevent Users from Changing their Role

a4bello:
## Implementation Description
A follow-up to https://github.com/uwblueprint/rowan-house/issues/136. A logged-in user should not be able to change their own role. There is a frontend and backend implementation to this ticket. The frontend implementation ensures that the edit role button in the `Manage Users` page does not appear when the logged-in user is searched for. The backend implementation acts as a sanity check to ensure a similar functionality. 

## Screenshots
image
image

- Ensure that if a logged-in user searches their name on the  Manage Users page, the edit button does not appear in the user card that appears on the page.
- Logged-in admin users should be able to change the user role of other users on the application. 

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -54,8 +54,13 @@ const executableSchema = makeExecutableSchema({
 const authorizedByAllRoles = () =>
   isAuthorizedByRole(new Set(["User", "Admin", "Staff"]));
 const authorizedByAdmin = () => isAuthorizedByRole(new Set(["Admin"]));
-const authorizeRoleChange = (id: string) => { return and(isAuthorizedToChangeRole(id), isAuthorizedByRole(new Set(["Admin"])))};
-//and(isAuthorizedToChangeRole(id), isAuthorizedByRole(new Set(["Admin"])))
+const authorizeRoleChange = (id: string) => {
+  return and(
+    isAuthorizedToChangeRole(id),
+    isAuthorizedByRole(new Set(["Admin"])),
+  );
+};
+// and(isAuthorizedToChangeRole(id), isAuthorizedByRole(new Set(["Admin"])))
 
 const graphQLMiddlewares = {
   Query: {
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -60,7 +60,6 @@ const authorizeRoleChange = (id: string) => {
     isAuthorizedByRole(new Set(["Admin"])),
   );
 };
-// and(isAuthorizedToChangeRole(id), isAuthorizedByRole(new Set(["Admin"])))
 
 const graphQLMiddlewares = {
   Query: {
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -6,6 +6,8 @@ import {
   isAuthorizedByEmail,
   isAuthorizedByRole,
   isAuthorizedByUserId,
+  isAuthorizedToChangeRole,
+  and,
 } from "../middlewares/auth";
 import authResolvers from "./resolvers/authResolvers";
 import authType from "./types/authType";
@@ -52,6 +54,8 @@ const executableSchema = makeExecutableSchema({
 const authorizedByAllRoles = () =>
   isAuthorizedByRole(new Set(["User", "Admin", "Staff"]));
 const authorizedByAdmin = () => isAuthorizedByRole(new Set(["Admin"]));
+const authorizeRoleChange = (id: string) => { return and(isAuthorizedToChangeRole(id), isAuthorizedByRole(new Set(["Admin"])))};
+//and(isAuthorizedToChangeRole(id), isAuthorizedByRole(new Set(["Admin"])))
 
 const graphQLMiddlewares = {
   Query: {
@@ -75,7 +79,7 @@ const graphQLMiddlewares = {
     deleteEntity: authorizedByAllRoles(),
     createUser: authorizedByAdmin(),
     updateUser: authorizedByAdmin(),
-    updateUserRole: authorizedByAdmin(),
+    updateUserRole: authorizeRoleChange("id"),
     deleteUserById: authorizedByAdmin(),
     deleteUserByEmail: authorizedByAdmin(),
     logout: isAuthorizedByUserId("userId"),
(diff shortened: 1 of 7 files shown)


backend/graphql/index.ts
```diff
@@ -1,89 +1,97 @@
...
+  ),
+});
+
+const authorizedByAllRoles = () =>
+  isAuthorizedByRole(new Set(["User", "Admin", "Staff"]));
+const authorizedByAdmin = () => isAuthorizedByRole(new Set(["Admin"]));
+const authorizeRoleChange = (id: string) => {
+  return and(
+    isAuthorizedToChangeRole(id),
+    isAuthorizedByRole(new Set(["Admin"])),
```
jfdoming, line 60:
```suggestion
    authorizedByAdmin(),
```

backend/middlewares/auth.ts
```diff
@@ -21,6 +30,81 @@ export const getAccessToken = (req: Request): string | null => {
...
+  if ("resolve" in resolver) {
+    return resolver.resolve?.(resolve, parent, args, context, info);
+  }
+  if (typeof resolver === "function") {
+    return resolver(resolve, parent, args, context, info);
+  }
+};
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
jfdoming, line 53: Instead of this, why not just set the return type to an `IMiddlewareFunction<any, ExpressContext>`?

backend/middlewares/auth.ts
```diff
@@ -21,6 +30,81 @@ export const getAccessToken = (req: Request): string | null => {
   return null;
 };
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
jfdoming, line 34: Would also be good to use explicit return types here.

backend/middlewares/auth.ts
```diff
@@ -21,6 +30,81 @@ export const getAccessToken = (req: Request): string | null => {
...
+    return resolver.resolve?.(resolve, parent, args, context, info);
+  }
+  if (typeof resolver === "function") {
+    return resolver(resolve, parent, args, context, info);
+  }
+};
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+export const and = (
```
jfdoming, line 54: Maybe worth moving this to its own/a different file since it applies more generally?

backend/middlewares/auth.ts
```diff
@@ -21,6 +30,81 @@ export const getAccessToken = (req: Request): string | null => {
...
+    const authorized =
+      accessToken &&
+      (await authService.isAuthorizedToChangeRole(
+        accessToken,
+        args[userIdField],
+      ));
+
+    if (!authorized) {
+      throw new AuthenticationError(
+        "Failed authentication and/or authorization by userId",
```
jfdoming, line 100: Could we change this error message to be something more descriptive? Something like "User ID cannot change its own role" or something.

backend/services/implementations/authService.ts
```diff
@@ -163,6 +163,30 @@ class AuthService implements IAuthService {
...
+      const tokenUserId = await this.userService.getUserIdByAuthId(
+        decodedIdToken.uid,
+      );
+
+      const firebaseUser = await firebaseAdmin
+        .auth()
+        .getUser(decodedIdToken.uid);
+
+      return (
+        firebaseUser.emailVerified && String(tokenUserId) !== requestedUserId
```
jfdoming, line 183: I don't think checking the `emailVerified` field is necessary here, since it can already be checked by `isAuthorizedByRole`. (If you remove the check here, it makes sense to reverse the order of the checks in the `and` in `backend/graphql/index.ts` so that the check runs first.)

backend/services/implementations/authService.ts
```diff
@@ -163,6 +163,30 @@ class AuthService implements IAuthService {
     }
   }
 
+  async isAuthorizedToChangeRole(
```
jfdoming, line 166: Could we change this to something more explicit about what it's doing, like `isActiveUserSameAsId` (and invert the logic)?

frontend/src/components/admin/UserCard.tsx
```diff
@@ -100,9 +102,11 @@ const UserCard = ({
     >
       <VStack align="center" spacing={8} ml="60px" mr="132px">
         <DefaultUserIcon initials={getInitials(firstName, lastName)} />
-        <Button onClick={onOpen} variant="outline-md">
-          Edit Role
-        </Button>
+        {authenticatedUser?.email !== email && (
```
jfdoming, line 105: 👍

jfdoming (changes requested):
Left a few comments, thanks for working on this! Feel free to push back if any of these things feel onerous.

backend/services/interfaces/authService.ts
```diff
@@ -50,7 +50,7 @@ interface IAuthService {
    * @returns true if token is valid and authorized; and if the id does not match
    * current logged in user's id, false otherwise
    */
-  isAuthorizedToChangeRole(
+  iDNotSameasActiveUser(
```
jfdoming, line 53: Suggestion:

```suggestion
  idNotSameasActiveUser(
```

backend/middlewares/utils/combinatorsUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+  parent: any,
+  args: { [key: string]: any },
+  context: ExpressContext | undefined,
+  info: GraphQLResolveInfo | undefined,
+) => any;
+
+export const callResolver = async (
+  resolver: IMiddlewareFunction<any, ExpressContext>,
+  resolve: Resolve,
+  parent: any,
```
jfdoming, line 16: Is this lint error fixable?

jfdoming (commented):
Nice refactors!

backend/services/interfaces/authService.ts
a4bello: Done

backend/middlewares/utils/combinatorsUtils.ts
a4bello: Yes, Done!

jfdoming (approved):

PR #188, opened 2022-07-05
Update Lesson Name Modal

r-aya:
## Implementation Description
Created a functioning modal for updating lesson names via the module editor.

## Screenshots
Screenshot 2022-07-04 at 10 12 55 PM

- Pre-existing lesson name updating to new name after confirming on modal
- Working error-handling for new lesson name input
- Name changes are saving

## Checklist


frontend/src/components/module-editor/LessonItem.tsx
```diff
@@ -58,7 +101,29 @@ const LessonItem = ({
...
+            <Modal
+              header="Edit lesson title"
+              isOpen={isOpen}
+              onConfirm={() => updateLessonName(title)}
+              onCancel={() => {
+                resetState();
+                onClose();
+              }}
+            >
+              <TextInput
```
jfdoming, line 116: Would be good to specify the `defaultValue` prop so the field is initialized with the current title. If you do this, it would also be good to _not_ reset the `title` state so that the default state still works on subsequent modal displays.

frontend/src/components/module-editor/LessonItem.tsx
```diff
@@ -16,6 +21,44 @@ const LessonItem = ({
...
+
+  const updateLessonName = (lessonTitle: string) => {
+    if (title) {
+      dispatch({
+        type: "update-lesson",
+        value: {
+          course: courseID,
+          module: course.modules[moduleID].id,
+          title: lessonTitle,
+          content: [],
```
jfdoming, line 51: Unfortunately, dispatching the action like this clears the lesson content, which is something we definitely don't want. Could you create a new action type instead which only updates the title?

jfdoming (changes requested):
Nice work! A few comments.

jfdoming (approved):

PR #206, opened 2022-07-19
[RHS-190]: Login Flow Improvements

a4bello:
## Implementation Description
The PR seeks to improve the login and signup flow. Refer to the ticket to see the list of things to fix for both pages. The login page now includes features like autofocus, autofill and error messages on login failure. The same also extends to the signup page. Previously, the signup page is in an error state by default, however, with this PR this only happens when a submission happens and there is actually an error. 

## Screenshots
Visual depiction of PR
image
image
image
image
image
image

- Reviewers should ensure all the features/implementations work as expected and as described in the ticket.

## Testing Procedure
   - What ways could the input be wrong? Invalid Email / Password
   - Is it secure? Could nefarious users access/break it? N/A

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/APIClients/AuthAPIClient.ts
+++ b/frontend/src/APIClients/AuthAPIClient.ts
@@ -35,7 +35,7 @@ const login = async (
     }
   } catch (e: unknown) {
     // eslint-disable-next-line no-alert
-    window.alert("Failed to login");
+    console.error("Failed to login");
   }
   return user;
 };
@@ -74,7 +74,7 @@ const register = async (
     }
   } catch (e: unknown) {
     // eslint-disable-next-line no-alert
-    window.alert("Failed to sign up");
+    console.error("Failed to sign up");
   }
   return user;
 };
--- a/frontend/src/APIClients/AuthAPIClient.ts
+++ b/frontend/src/APIClients/AuthAPIClient.ts
@@ -34,7 +34,7 @@ const login = async (
       localStorage.setItem(AUTHENTICATED_USER_KEY, JSON.stringify(user));
     }
   } catch (e: unknown) {
-    // eslint-disable-next-line no-alert
+    // eslint-disable-next-line no-console
     console.error("Failed to login");
   }
   return user;
@@ -73,7 +73,7 @@ const register = async (
       localStorage.setItem(AUTHENTICATED_USER_KEY, JSON.stringify(user));
     }
   } catch (e: unknown) {
-    // eslint-disable-next-line no-alert
+    // eslint-disable-next-line no-console
     console.error("Failed to sign up");
   }
   return user;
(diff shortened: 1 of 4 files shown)


frontend/src/components/auth/Login.tsx
```diff
@@ -146,22 +199,53 @@ const Login = (): React.ReactElement => {
...
+                ref={emailRef}
+                type="email"
+                value={email}
+                placeholder="you@rowanhouse.ca"
+                onChange={(event: ChangeEvent<HTMLInputElement>) =>
+                  setEmail(event.target.value)
+                }
+                marginBottom="3vh"
+                autofill
+                style={{ display: "none" }}
```
jfdoming, line 221: `display="none"`

frontend/src/components/auth/Login.tsx
```diff
@@ -48,16 +67,23 @@ const Login = (): React.ReactElement => {
...
+      setLoginFail(true);
+    }
     setAuthenticatedUser(user);
   };
 
   const onLogInClick = async () => {
     switch (loginState) {
       case LoginState.EnterEmail:
-        setLoginState(LoginState.EnterPassword);
+        if (emailRef.current?.validity.valid) {
```
jfdoming, line 79: If this is not true, this leaves no indication to the user why their validation failed. Could you add an extra error state where the email is incorrect, telling them that what they entered isn't a valid email?

frontend/src/components/auth/Login.tsx
```diff
@@ -101,22 +129,47 @@ const Login = (): React.ReactElement => {
...
+                Email address
+              </FormLabel>
               <Input
+                ref={emailRef}
                 type="email"
                 value={email}
                 placeholder="you@rowanhouse.ca"
-                onChange={(event) => setEmail(event.target.value)}
+                onChange={(event: ChangeEvent<HTMLInputElement>) =>
+                  setEmail(event.target.value)
```
jfdoming, line 146: If the user modifies their email here, we should reset the contents of the password field. This is so the browser doesn't think the user submitted an autofilled email.

frontend/src/components/auth/Signup.tsx
```diff
@@ -82,7 +88,7 @@ const Signup = (): React.ReactElement => {
             </Center>
             <FormControl
```
jfdoming, line 89: Could you change each of the form elements (label + input) to be in their own form control, instead of all being in the same one? I think this will fix some of the autofill issues we're experiencing still on this page.

jfdoming (changes requested):
Nice work! Some comments.

jfdoming (commented):
Awesome job! One last request (should be quick): would you mind removing the alerts from `AuthAPIClient.ts`? Now that we have the error messages built in they're unnecessary, we can just replace them with a `console.error` or something.

frontend/src/APIClients/AuthAPIClient.ts
```diff
@@ -35,7 +35,7 @@ const login = async (
     }
   } catch (e: unknown) {
     // eslint-disable-next-line no-alert
```
jfdoming, line 37: One last thing: could you update the eslint comments to reflect the console log instead of the alert?

jfdoming (changes requested):

jfdoming (approved):

PR #211, opened 2022-07-20
Fix missing type in signup flow

jfdoming:
## Implementation Description
Fix missed type in User -> Learner migration.

## Screenshots
N/A

- Can you find any other types (I couldn't)?

## Testing Procedure
Tested by logging in locally, it does seem to work now.

## Things to Note
This is why integration tests would be nice...

## Checklist


a4bello (commented):
I went through the code, and didn't see any instance of User. I think you're good to go 👍

a4bello (approved):

PR #214, opened 2022-07-25
Course progress backend

jfdoming:
## Implementation Description
- Update module deletion to assign to null instead of removing
- Add course/lesson progress models and service
- Add course/lesson progress resolvers

Highly recommend reviewing by commit!

## Screenshots
N/A

- Are the created endpoints useful?
- Are there use cases which are not handled by either the schema design or the query design?

## Testing Procedure
Checkout locally, use Apollo extension to manually query endpoints.

## Things to Note
N/A

## Checklist

### DIFF
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -97,6 +97,7 @@ const graphQLMiddlewares = {
     lessonById: authorizedByAllRoles(),
     lessons: authorizedByAllRoles(),
     courseProgress: isAuthorizedByUserId("userId"),
+    moduleProgress: isAuthorizedByUserId("userId"),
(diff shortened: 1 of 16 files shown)


draymond63 (commented):
`courseProgress` API returns data in an array. How does this data relate to the course IDs it was sent? I tried swapping the order of the course IDs in the request, but that didn't change the order of the data. I feel like the response should include either include the course IDs the data pertains to, or make the order matter.

`moduleProgress` API seems good.

`lessonProgress` API makes sense, although it only returns an object if a lesson has been marked as completed. That works, but it seems a little cumbersome to work with on the frontend side (although, I'm not sure about that)

draymond63 (approved):
Looks good to me!

Issue #129, opened 2022-05-13
Module Editor: Delete lessons


jfdoming:
## Description
We'd like to provide users the ability to delete lessons, along with the existing add/update actions they can perform. When the user attempts to delete the lesson by clicking the icon, a modal will appear to confirm their decision. 

Ideally, the lesson should be deleted immediately after confirmation. However, this is complicated, so for now we will say that the deletion will take place once the user saves their changes. 

## Goals
- [x] Show deletion modal when the lesson's delete button is pressed
- [ ] Update frontend state logic

### UI
The visuals for this ticket will be written in `SidebarModuleOverview.tsx`.  State will need to be added to keep track of the hover status, similar to that in `EditableContentBlock.tsx`. For the modal, use the `frontend/src/components/common/DeleteModal.tsx`.  

### State Logic
Once the user confirms they want to delete the lesson, this state has to be updated to match this. Due to the de-sync'd nature between the frontend state and the backend database, different things should occur depending on the current state.
1. If the lesson already exists in the database, then the `hasChanged` state value should add a `DELETE` value associated with the lesson ID.
  - This requires us to then use the `deleteLesson` mutation on a save. This is implemented in the `saveChanges` function within `frontend/src/components/module-editor/Sidebar.tsx`. It might work already, but you should make sure it does.
2. If the lesson was created and deleted in the same period (without saving), then this lesson only exists on the frontend. Thus, we can simply remove the `CREATE`'d lesson from the `hasChanged` state.
  - You'll need to update the `updateChangeStatus` in `frontend/src/reducers/ModuleEditorContextReducer.ts` to implement this logic.

PR #223, opened 2022-08-02
Implement Lesson Deletion [working]

molly-yu:
## Implementation Description
Implements lesson deletion: https://github.com/uwblueprint/rowan-house/issues/129
- Create and add delete modal
- Backend fixes for delete lesson mutation
- Frontend state logic update + fixes to reducer, saveChanges, changeStatus behavior
- add moduleIndex to actions to b/c we cannot fetch module by ID

## Screenshots
Visual depiction of PR
Screen Shot 2022-08-01 at 11 46 31 PM

- testing different combinations of deleting/creating/saving lessons + page reloading

## Testing Procedure

## Things to Note

## Checklist


frontend/src/components/module-editor/SideBarModuleOverview.tsx
```diff
@@ -9,15 +9,24 @@ import { Modal } from "../common/Modal";
...
+
+enum ModalType {
+  EDIT = "edit",
+  DELETE = "delete",
+  CREATE_LESSON = "create-lesson",
+}
 
 const SideBarModuleOverview = (): React.ReactElement => {
   const context = useContext(EditorContext);
   const { isOpen, onOpen, onClose } = useDisclosure();
```
jfdoming, line 22: Instead of adding a `ModalType` enum, it would be better to have three `useDisclosure` calls, that way the modal components don't have to mount/unmount when the type changes.

jfdoming (changes requested):
Thanks for putting in so much work for this one!! Left a few comments but overall looks good.

jfdoming:
Also unfortunately looks like there are some merge conflicts :disappointed:

jfdoming (approved):

PR #227, opened 2022-08-06
Make course/lesson info API routes public

jfdoming:
## Implementation Description
- Make course info API routes public
- Add new public route returning minimal lesson information

The commits are disjoint so I'd recommend reviewing by commit.

## Screenshots
N/A

- Is the logic secure?
- Does the new route cover all remaining use cases for public lesson info?

## Testing Procedure
Checkout branch and try GQL endpoints as a non-logged-in user.

## Things to Note
None

## Checklist

### DIFF
--- a/backend/middlewares/auth.ts
+++ b/backend/middlewares/auth.ts
@@ -11,14 +11,14 @@ const authService: IAuthService = new AuthService(new UserService());
 
 export const getAccessToken = (req: Request): string | null => {
   const [scheme, token] = req.headers.authorization?.split(" ") || [];
-  if (scheme.toLowerCase() === "bearer") {
+  if (scheme?.toLowerCase() === "bearer") {
     return token === "null" ? null : token;
   }
   return null;
 };
 
-/* Determine if the user whose role is to be changed does not match the currently 
-logged in user. 
+/* Determine if the user whose role is to be changed does not match the currently
+logged in user.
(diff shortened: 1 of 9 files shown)


backend/middlewares/auth.ts
```diff
@@ -10,13 +10,9 @@ import { Role } from "../types";
...
 export const getAccessToken = (req: Request): string | null => {
-  const authHeaderParts = req.headers.authorization?.split(" ");
-  if (
-    authHeaderParts &&
-    authHeaderParts.length >= 2 &&
-    authHeaderParts[0].toLowerCase() === "bearer"
-  ) {
-    return authHeaderParts[1];
+  const [scheme, token] = req.headers.authorization?.split(" ") || [];
+  if (scheme.toLowerCase() === "bearer") {
```
draymond63, line 14: If `scheme` is null does this work?

draymond63 (approved):
I think this all makes sense. Just one question, probably doesn't need to be resolved.

backend/middlewares/auth.ts
jfdoming: Ah good point. I don't think it would ever come up practically but I'll fix it nonetheless. Really surprised Typescript didn't find this one though...

draymond63 (approved):

PR #234, opened 2022-08-26
Fix mismatched endpoint name for Signup

jfdoming:
## Implementation Description
We were refreshing the auth token for signup endpoints due to a naming issue, which caused auth errors. This PR migrates `SignUp` -> `Signup_Register`.

## Screenshots
N/A

- 

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/APIClients/BaseAPIClient.ts
+++ b/frontend/src/APIClients/BaseAPIClient.ts
@@ -73,7 +73,7 @@ const accessTokenInjectionLink = new ApolloLink(
 
 const refreshDirectionalLink = new RetryLink().split(
   (operation) =>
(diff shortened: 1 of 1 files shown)


Tayef-Shah (approved):

draymond63:
I have no insight into this, so I'm not much help here. Lmk and I can do a deep dive into this, but otherwise this is your call

jfdoming:
> I have no insight into this, so I'm not much help here. Lmk and I can do a deep dive into this, but otherwise this is your call

At the very least, this fixes an issue I was encountering locally. 😛 I can investigate more deeply if other issues persist.

PR #245, opened 2022-09-05
Module Completion Page: Up Next

mslwang:
## Implementation Description
We want to display a card on the module completion page indicating the up-next module.

## Screenshots
image
image
image

- Testing user flow for viewing courses
- Check for different stages of module completion: (Not started, In Progress, Complete)

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/components/module-viewer/UpNext.tsx
+++ b/frontend/src/components/module-viewer/UpNext.tsx
@@ -1,5 +1,5 @@
 import { useLazyQuery } from "@apollo/client";
-import { Heading } from "@chakra-ui/react";
+import { Heading, Spinner } from "@chakra-ui/react";
 import React, { useContext, useState, useEffect } from "react";
 import { GET_MODULE_PROGRESS } from "../../APIClients/queries/ProgressQueries";
 import {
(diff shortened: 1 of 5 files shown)


mslwang:
Remaining todos:
- Fix error with re-direction after clicking Up Next card when subsequent lesson is present
image
- Fix issue where Module Completion page remains after clicking Up Next card if next module does not have any lessons. 

The current assumption elsewhere seems to be that `moduleProgress` length should be the name same as `modules` length, but the following screenshot says otherwise 
image
So if we need to account for people who go to their module page from elsewhere (ie. protection for if you jump from module 3 to module 5 without doing module 4 from the url, probably out of scope of this ticket though), the moduleProgress array won't work correctly since the indices need to line up.

mslwang:
Note: @a4bello, I saw this issue, but I doubt it's one that we need to focus on. When course has been edited after someone has already completed the course, the course completion does not get updated, which messes with `View Course` directing you to the correct module.

https://user-images.githubusercontent.com/53355975/188486851-c1e3bfd1-581f-4a0a-930d-1a1b2ecef540.mov

frontend/src/components/module-viewer/UpNext.tsx
```diff
@@ -0,0 +1,88 @@
...
+  useEffect(() => {
+    if (authenticatedUser) {
+      getModuleProgress({
+        variables: {
+          userId: authenticatedUser.id,
+          courseId: courseID,
+        },
+      });
+    }
+  }, [authenticatedUser, setAuthenticatedUser, getModuleProgress, courseID]);
```
jfdoming, line 47: Doesn't look like setAuthenticatedUser is actually used?

jfdoming (approved):

frontend/src/components/module-viewer/UpNext.tsx
mslwang: Fixed

jfdoming (approved):

jfdoming (approved):

Issue #137, opened 2022-05-17
Dashboard: download data modal


draymond63:
## Summary
The client needs to have information about people's location for funding purposes. The backend functionality has already been created to retrieve this information, but we need a frontend UI to let the user interact with it. The goal of this ticket is to create a modal that will allow the user to specify a date range, and then download the resulting info to their filesystem for safe keeping. 

![image.png](https://images.zenhubusercontent.com/6053e4ca8db077e41d15aa4a/8644d38f-cf05-4b97-a765-6d7b5f75c0ed)

For the full design/flow, see the [Figma](https://www.figma.com/file/Yb3MvhfQAhfCjLg8cnNCoo/User-Management?node-id=525%3A607).

### Additional Notes
- Defaults to the "Download All" option
- If a date is edited, automatically switch the the "Date Range" option
- If the dates provided are invalid (i.e. end is before start, year is not valid), show error
- Date range can be open-ended/unspecified, resulting in an open-ended query
- Year textboxes should only accept numerical entries

## Goals
- [x] Create modal that opens when the "Download Data" button is pressed
- [x] Add date picker to the modal
- [x] Call the `userCountByTown` query and save the info as a CSV to the filesystem

PR #250, opened 2022-09-08
add modal for downloading user stats

Tayef-Shah:
## Implementation Description
Added a modal for downloading and added the date pickers as a separate component. The apollo hook used is the `useLazyQuery` hook since we only want to query the data when the user clicks download.

## Screenshots
- Downloading is disabled until a complete date range is provided OR `download all data` is selected
- An invalid date range shows an error
- Non numeric characters do not appear in the year input

Screen Shot 2022-09-28 at 10 45 12 PM
Screen Shot 2022-09-28 at 10 45 20 PM
Screen Shot 2022-09-28 at 10 45 39 PM

- Testing

## Testing Procedure
- Follow the issue #137 for info on what the expected behaviour should be.
- Test that download button is only disabled if you are selecting specific date ranges, but have not provided all date range info
- Test that an error shows up if the start date is later than the end date
- Test that closing the modal and returning to it will start in a fresh state

## Things to Note
> Note: Instead of an x button to exit the modal, there's a cancel button. This was done so that the existing modal component could be reused. Hopefully this doesn't make the user experience significantly different.

## Checklist

### DIFF
--- a/frontend/src/components/admin/DatePicker.tsx
+++ b/frontend/src/components/admin/DatePicker.tsx
@@ -1,0 +1,83 @@
+import React, { useState } from "react";
+import { Text, Stack, Select, Input } from "@chakra-ui/react";
+
+type DatePickerProps = {
+  label: string;
+  day: string;
+  setDate: (date: string) => void;
+};
+
+const DatePicker = ({
+  label,
+  day,
+  setDate,
+}: DatePickerProps): React.ReactElement => {
+  enum Months {
+    January = "01",
+    February = "02",
+    March = "03",
+    April = "04",
+    May = "05",
+    June = "06",
+    July = "07",
+    August = "08",
+    September = "09",
+    October = "10",
+    November = "11",
+    December = "12",
+  }
+
+  const [month, setMonth] = useState("");
+  const [year, setYear] = useState("");
+  const [invalidYear, setInvalidYear] = useState(false);
+
+  const updateDate = () => {
+    if (month !== "" && year !== "") {
+      setDate(`${year}/${Months[month as keyof typeof Months]}/${day}`);
+    }
+  };
+
+  const handleYearInput = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const inputYear = event.target.value;
+    setYear(inputYear);
+    setInvalidYear(!Number.isInteger(Number(inputYear)));
+    if (!invalidYear) {
+      updateDate();
+    }
+  };
+
+  const handleMonthInput = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const inputMonth = event.target.value;
+    setMonth(inputMonth);
+    updateDate();
+  };
+
+  return (
+    <Stack direction="column">
+      <Text>{label}</Text>
+      <Stack direction="row">
+        <Select
+          placeholder="Month"
+          size="sm"
+          value={month}
+          onChange={handleMonthInput}
+        >
+          {Object.keys(Months).map((m) => (
+            <option key={m} value={m}>
+              {m}
+            </option>
+          ))}
+        </Select>
+        <Input
+          isInvalid={invalidYear}
+          placeholder="Year"
+          value={year}
+          onChange={handleYearInput}
+          size="sm"
+        />
+      </Stack>
+    </Stack>
+  );
+};
+
+export default DatePicker;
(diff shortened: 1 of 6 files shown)


frontend/src/components/admin/DatePicker.tsx
```diff
@@ -0,0 +1,73 @@
...
+          value={Object.keys(Months)[Number(month) - 1]}
+          onChange={handleMonthInput}
+        >
+          {Object.keys(Months).map((m) => (
+            <option key={m} value={m}>
+              {m}
+            </option>
+          ))}
+        </Select>
+        <Input
```
jfdoming, line 61: Might be nice to switch this to a numeric input type on mobile: https://caniuse.com/input-inputmode. Not high priority though if it looks complicated, admins will most likely use desktop most of the time.

frontend/src/components/admin/DatePicker.tsx
```diff
@@ -0,0 +1,73 @@
...
+    setYear(event.target.value.replace(/\D/g, ""));
+  };
+
+  const handleMonthInput = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setMonth(Months[event.target.value as keyof typeof Months]);
+  };
+
+  return (
+    <Stack direction="column">
+      <Text>{label}</Text>
```
jfdoming, line 46: Could you use the FormLabel and FormControl elements for this by any chance? It will have better accessibility properties and will make focusing the inputs easier.

frontend/src/components/admin/DownloadDataModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const [startMonth, setStartMonth] = useState("");
+  const [startYear, setStartYear] = useState("");
+  const [endMonth, setEndMonth] = useState("");
+  const [endYear, setEndYear] = useState("");
+  const [getCounts, { loading }] = useLazyQuery(GET_USER_COUNT_BY_TOWN);
+  const [downloadEnabled, setDownloadEnabled] = useState(true);
+  const [rangeInvalid, setRangeInvalid] = useState(false);
+
+  useEffect(() => {
+    if (radioOption === "2") {
```
jfdoming, line 28: Instead of doing this in a useEffect, it would be good to put it directly in the onChange handler for the controls (perhaps refactored into a helper?). That way we can avoid the extra rerender needed to actually update this state.

frontend/src/components/admin/DownloadDataModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+    borderWidth: "1px",
+    w: "100%",
+    padding: "1rem",
+  };
+  const radioProps = { spacing: "1rem", w: "100%" };
+  return (
+    <Modal
+      size="2xl"
+      header="Pick timeframe to download data"
+      confirmText={loading ? "Downloading..." : "Download"}
```
jfdoming, line 83: Chakra buttons have a dedicated loading state. What do you think about adding a new prop to the modal component and using that instead? (i.e., is it worth the effort for a more consistent feel?))

Tayef-Shah: Yeah I'm down, it's a small change

frontend/src/components/admin/DatePicker.tsx
Tayef-Shah: Sure, I'm not really sure what props I should add for accessibility though. Should I be adding aria-labels?

frontend/src/components/admin/DownloadDataModal.tsx
Tayef-Shah: Finding this to be a little tricky to implement. At the moment it's not validating exactly when the input is valid, but rather requires one extra change to the input for the validation to run. I'm guessing this is because updating the state is asynchronous and the validation happens before the state is updated. Any ideas on how to fix this?

frontend/src/components/admin/DatePicker.tsx
jfdoming: Should be fine to just use FormLabel and FormControl, Chakra components should already have any necessary aria- attributes specified.

frontend/src/components/admin/DownloadDataModal.tsx
jfdoming: I guess if it works it's not the end of the world to keep it this way, if the fix is complicated. FF to skip this one.

jfdoming (approved):

frontend/src/components/admin/DownloadDataModal.tsx
Tayef-Shah: FF 💀

Issue #216, opened 2022-07-31
Course Browsing Page: Course Filters


jfdoming:
## Summary
We want to make courses filterable based on a user's progress.

## Goals
- [ ] Fetch data from the GQL `GET_COURSE_PROGRESS` endpoint.
- [ ] Use the filter buttons to filter a user's progress.

## Designs
![image](https://user-images.githubusercontent.com/9922514/182005748-abf4a0f4-5a8c-46d2-ad89-63370bc7b492.png)
https://www.figma.com/file/hVyZkprXKN6myQuaIvRqfr/Learner-Views?node-id=227%3A3662

## Implementation
See above.

PR #255, opened 2022-09-15
Can filter by progress

smilesthesad:
## Implementation Description
Clicking the top tabs allows you to filter by started, in progress, and completed. Addresses https://github.com/uwblueprint/rowan-house/issues/216

## Screenshots
Visual depiction of PR

https://user-images.githubusercontent.com/36215359/190315031-3c7eb6c9-9b67-40bb-bc3f-be2c1025add0.mov

## Testing Procedure
Create 3 courses, and as a learner start one of them and finish another.
- Observe courses that have not been started are in not started
- Observe courses that have been started are in progress
- Observe courses than have been completed are completed
 

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/components/pages/Default.tsx
+++ b/frontend/src/components/pages/Default.tsx
@@ -11,7 +11,6 @@ import {
 } from "../../APIClients/queries/CourseQueries";
 import { CourseResponse } from "../../APIClients/types/CourseClientTypes";
 import { GET_COURSE_PROGRESS } from "../../APIClients/queries/ProgressQueries";
-import { CourseType } from "../../types/ModuleEditorTypes";
 import { CourseProgressResponse } from "../../APIClients/types/ProgressClientTypes";
 
 const Default = (): React.ReactElement => {
@@ -20,8 +19,8 @@ const Default = (): React.ReactElement => {
   const [courses, setCourses] = useState([]);
 
   const [courseProgresses, setCourseProgresses] = useState<
-    Array<CourseProgressResponse>
-  >([]);
+    Map<string, string>
+  >();
 
   const [selectedTab, setSelectedTab] = useState<string>(TAB_NAMES[0]);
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/pages/Default.tsx
```diff
@@ -60,9 +110,12 @@ const Default = (): React.ReactElement => {
...
           spacing={5}
         >
-          {courses?.map((course: CourseResponse) => (
-            <CourseCard key={course.id} course={course} />
-          ))}
+          {courses?.map((course: CourseResponse) => {
+            if (checkFilter(selectedTab, course.id)) {
+              return <CourseCard key={course.id} course={course} />;
+            }
+            return <></>;
```
jfdoming, line 117: Might be good to return `null` instead to avoid creating extra nodes.

frontend/src/components/pages/Default.tsx
```diff
@@ -40,8 +59,39 @@ const Default = (): React.ReactElement => {
...
 
+  const checkFilter = (tab_name: string, courseId: string): boolean => {
+    if (tab_name === "All Courses") {
+      return true;
+    }
+    if (!courseProgresses) {
+      return true;
+    }
+    const courseObj = courseProgresses.find(
+      (course) => course.courseId === courseId,
```
jfdoming, line 74: This is kind of inefficient (quadratic time), which you could work around by making courseProgresses a map instead. However, the scale might be too small for this to matter so I'm wondering your thoughts.

smilesthesad: my original thought process was because the scale is pretty small it was okay, but I think ill refactor anyways

frontend/src/components/pages/Default.tsx
draymond63: I believe react fragments don't result in any dom elements. So, this should be equivalent, right?

jfdoming: Technically I think it creates a virtual DOM node in React, but it probably doesn't matter practically, I guess.

jfdoming (approved):

PR #256, opened 2022-09-22
added progress tags

smilesthesad:
## Implementation Description
Added Progress Tags in the Course Browsing Page

## Screenshots
image

- Ensure that the progress tag matches the progress of the actual course
- Ensure that when a user is not logged in none of the progress tags show up

## Testing Procedure
- Create 3+ courses
- Start at least one of them, complete another, and leave at least one unfinished
- Ensure that when you go to the course view page, they all have the right progress attached to them

## Checklist


frontend/src/components/learner/browser/CourseCard.tsx
```diff
@@ -25,6 +42,52 @@ const CourseCard = ({ course }: CourseCardProps): React.ReactElement => {
     ) || 0;
   const expanded = focused || hovered;
 
+  const StatusTag = () => {
```
jfdoming, line 45: Might be better to pass `status` as a prop, that way this component can be defined out of line.

frontend/src/components/learner/browser/CourseCard.tsx
```diff
@@ -42,11 +42,11 @@ const CourseCard = ({
     ) || 0;
   const expanded = focused || hovered;
 
-  const StatusTag = () => {
-    if (!status) {
+  const StatusTag = (courseProgress: string | null) => {
```
jfdoming, line 45: I'm fine with this not being a component (though ideally it would be) but it would definitely be nice to extract this out of line to the top level, or even to a separate file.

jfdoming (approved):

PR #266, opened 2022-10-08
Quiz component

draymond63:
## Implementation Description
Implements the quiz component, capable of multi-select and single select knowledge checks. Doesn't follow the exact design but it's pretty close :) https://www.figma.com/file/Dimm13cjrpYU5LkcrV6VxF/Course-Editor?node-id=1574%3A11848

## Screenshots
image
image
image

## Checklist


backend/middlewares/validators/util.ts
```diff
@@ -101,6 +102,15 @@ export const validateContent = (content: ContentBlock): boolean => {
         flipCard.content.cards !== null && flipCard.content.cards.length >= 1
       );
     }
+    case "quiz": {
+      const quiz = content as QuizBlock;
+      return (
+        quiz.content.question !== null &&
+        ["MS", "MC"].includes(quiz.content.type) &&
```
jfdoming, line 109: Nit: it seems a little unclear what these stand for. Maybe use longer strings to help with understanding?

frontend/src/components/common/checkboxes/CheckBox.css
```diff
@@ -0,0 +1,55 @@
+/* Used only for Quiz component */
```
jfdoming, line 1: I'd prefer to put this inline in the component, but if it's easier to keep it this way, would you be able to make the class names themselves a little more checkbox-specific? That way it's clear what they're for from reading the HTML.

frontend/src/components/common/checkboxes/CheckBox.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onClick,
+  isSelected = false,
+  radio = false,
+  size = "22px",
+  selectedColor = "#94D969",
+  unselectedColor = "#ddd",
+  disabled = false,
+}: CheckBoxParams): React.ReactElement => {
+  return (
+    <div className="container" onClick={disabled ? undefined : onClick}>
```
jfdoming, line 27: Can you double check if there are aria- attributes we should add for a11y? There may not be but I just wanted to make sure we're thinking about it.

frontend/src/components/common/checkboxes/CheckBoxes.tsx
```diff
@@ -0,0 +1,86 @@
...
+  setStatus: (statuses: boolean[]) => void;
+  statuses: boolean[];
+  multiSelect?: boolean;
+  padding?: number;
+  borderStyling?: (index: number) => string | undefined;
+  backgroundStyling?: (index: number) => string | undefined;
+  children: React.ReactElement[];
+}
+
+const CheckBoxes = ({
```
jfdoming, line 16: Broadly speaking, what is the benefit of a custom checkbox component over using checkboxes and radio buttons? Not opposed to it either way, but curious. Also if we do that, maybe we should rename this something more descriptive like `CheckBoxGroup` or something.

frontend/src/components/module-viewer/content/blocks/QuizBlock.tsx
```diff
@@ -0,0 +1,83 @@
...
+      paddingRight="4rem"
+      align="left"
+      borderTopColor="brand.royal"
+      borderTopWidth="10px"
+      borderRadius="8px"
+      boxShadow="xl"
+    >
+      <Flex align="center">
+        <QuizIcon />
+        <Text pl="8px" variant="heading">Knowledge Check</Text>
```
jfdoming, line 54: Linter? Also can probably use a HStack for automatic inter-element padding instead.

frontend/src/components/module-viewer/content/blocks/QuizBlock.tsx
```diff
@@ -0,0 +1,83 @@
...
+        backgroundStyling={backgroundStyling}
+        disabled={isCompleted}
+      >
+        {content.choices.map(({ answer }, i) => (
+          <Text key={i}>{answer}</Text>
+        ))}
+      </CheckBoxes>
+      {!isCompleted && (
+        <Button
+          w="fit-content"
```
jfdoming, line 72: Is this necessary?

draymond63: Weirdly yeah. It defaults to fill the parent, which is strange.

frontend/src/theme/foundations/colors.js
```diff
@@ -1,7 +1,7 @@
 export default {
   brand: {
     royal: "#724A96",
-    purple: "#CC99FF",
+    purple: "#724A96",
```
jfdoming, line 4: This change might impact a lot of places, is that fine?

draymond63: Whoops that was a mistake. I reverted it.

frontend/src/components/common/checkboxes/CheckBoxes.tsx
draymond63: More flexibility, but I've changed the implementation because it's probably better you use the libraries.

PR #267, opened 2022-10-09
Remove starter code

draymond63:
## Implementation Description
Removed all mention of "entity" endpoints and uses. Please go through and make sure I didn't delete any files that might be necessary.


frontend/src/utils/FileUtils.ts
```diff
@@ -1,9 +0,0 @@
-// eslint-disable-next-line import/prefer-default-export
```
jfdoming, line 1: Can you double check this file is not used? Just thinking about CSV downloads.

jfdoming (changes requested):
Could you run the linter for these changes? Also, I'm wondering if there are libraries we can remove as well (could be beneficial in the sense of reducing bundle size).

frontend/src/utils/FileUtils.ts
draymond63: Searching for `downloadFile` in the code, I couldn't find any example of it. There's a whole `CSVUtils` file with the function `generateCSV` that handles CSV download.

draymond63:
> Could you run the linter for these changes?

I just did and there didn't seem to be any problems or changes made by the linter. Did you notice anything wrong?

frontend/src/utils/FileUtils.ts
jfdoming: Great, thanks!

jfdoming:
> > Could you run the linter for these changes?
> 
> I just did and there didn't seem to be any problems or changes made by the linter. Did you notice anything wrong?

I see a comment on the PR in App.tsx, but maybe it's stale.

jfdoming (approved):

PR #269, opened 2022-10-11
Display module images everywhere

jfdoming:
## Implementation Description
Display module images everywhere

## Screenshots
Visual depiction of PR

- 

## Testing Procedure

## Things to Note

## Checklist

### DIFF
--- a/backend/services/implementations/courseService.ts
+++ b/backend/services/implementations/courseService.ts
@@ -7,6 +7,7 @@ import {
   CourseResponseDTO,
   ICourseService,
 } from "../interfaces/ICourseService";
+import IImageUploadService from "../interfaces/IImageUploadService";
 import MgCourse, {
   Course,
   CourseVisibilityAttributes,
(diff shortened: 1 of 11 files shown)


backend/services/implementations/courseService.ts
```diff
@@ -34,7 +35,48 @@ const publishedModulesOnlyQuery = [
...
 
 class CourseService implements ICourseService {
+  imageService: IImageUploadService;
+
+  constructor(imageService: IImageUploadService) {
+    this.imageService = imageService;
+  }
+
   /* eslint-disable class-methods-use-this */
+  private async hydrateImages(
```
draymond63, line 45: Just so I understand, what is the point of the backend hydration? Are you downloading the image in the backend so all the frontend has to do is use the attribute?

backend/services/implementations/courseService.ts
```diff
@@ -34,7 +35,48 @@ const publishedModulesOnlyQuery = [
...
+    course: CourseResponseDTO,
+  ): Promise<CourseResponseDTO> {
+    const coursePreviewImage = course.image
+      ? await this.imageService.download(course.image)
+      : course.previewImage;
```
draymond63, line 50: Do courses have `image` and `previewImage` fields now? Is this new?

backend/services/implementations/courseService.ts
```diff
@@ -34,7 +35,48 @@ const publishedModulesOnlyQuery = [
...
+    const newModules = await Promise.all(
+      course.modules.map(async (module) => {
+        const modulePreviewImage = module?.image
+          ? await this.imageService.download(module.image)
+          : module?.previewImage;
```
draymond63, line 55: I know when I made the data model, I had put in a module `image` and `previewImage` property. Obviously I've never made use of them, but is there a meaningful distinction between the two? It seems like previewImage is being used as a backup?

draymond63 (commented):
Just some general questions for my knowledge's sake :)

backend/services/implementations/courseService.ts
jfdoming: Kind of. Firebase uses temporary signed URLs based on a "file path" to access images, but these expire every hour so we can't store them. Each time we want to access an image, we have to recreate this URL (or use one from within the hour). So it's not downloading the image so much as getting the image URL.

draymond63 (approved):
Sweet! Looks good :)

PR #271, opened 2022-10-20
Editor niceties

draymond63:
## Implementation Description
- [x] Component kiosk closes a section when you click anywhere (should only close when you click the header)
- [x] Saving changes switches to latest completed lesson in editor (Treats admin as a learner)
- [x] Fix image errors
- [x] Padding between components in a lesson should look more natural
- [x] Text editor doesn't expand for longer paragraphs
- [x] Video in column component is too wide
- [x] Increase font size of text in flip cards
- [x] Standardize add/edit modals in module editor (Confirm vs Save, Edit vs Add copy) 
- [x] Match block component throws key error

### DIFF
--- a/frontend/src/APIClients/BaseAPIClient.ts
+++ b/frontend/src/APIClients/BaseAPIClient.ts
@@ -128,6 +128,9 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
   }
   // eslint-disable-next-line no-console
   if (networkError) console.log("[Network error]", networkError);
+  if (networkError?.message === "Error: Failed to refresh access token!") {
+    window.location.href = '/logout';
+  }
 });
 
 /*
--- a/frontend/src/APIClients/BaseAPIClient.ts
+++ b/frontend/src/APIClients/BaseAPIClient.ts
@@ -129,7 +129,7 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
   // eslint-disable-next-line no-console
   if (networkError) console.log("[Network error]", networkError);
   if (networkError?.message === "Error: Failed to refresh access token!") {
-    window.location.href = '/logout';
+    window.location.href = "/logout";
   }
 });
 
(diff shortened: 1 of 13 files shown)


frontend/src/components/module-viewer/SideBarContentKiosk.tsx
```diff
@@ -40,20 +40,24 @@ const SideBarContentKiosk = (): React.ReactElement => {
...
-                  return newState;
-                })
-              }
               borderTop="1px"
               borderColor="background.grey"
               padding="0.5rem"
             >
-              <Flex>
+              <Flex
+                _hover={{
```
jfdoming, line 48: Nit: you probably don't need to set this on hover, since cursors only apply on hover anyway.

frontend/src/components/module-viewer/BlockPreview.tsx
```diff
@@ -24,40 +24,40 @@ const BlockStyle = ({
   let contentIcon: React.ReactElement;
   switch (content.preview) {
     case "column.svg":
-      contentIcon = <ColumnIcon width="fit-content" height="fit-content" />;
+      contentIcon = <ColumnIcon width="100%" height="100%" />;
       break;
```
jfdoming, line 28: 🎉

frontend/src/APIClients/BaseAPIClient.ts
```diff
@@ -128,6 +128,9 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
   }
   // eslint-disable-next-line no-console
   if (networkError) console.log("[Network error]", networkError);
+  if (networkError === "Error: Failed to refresh access token!") {
+    location.href = '/logout';
```
jfdoming, line 132: Probably just need to disable this l'inter error, since we don't have a choice here.

jfdoming (commented):
Great! Left a few comments, and also noticed a few lint issues in general.

frontend/src/APIClients/BaseAPIClient.ts
draymond63: Yeah I updated it so I think it should be fixed. However, I have no idea how to test this except for waiting to see if the error occurs. Do you have any ideas how I can at least force this function to run, ideally with the debugger?

jfdoming (approved):

PR #273, opened 2022-10-21
add image folder

smilesthesad:
## Implementation Description
Check upload path and if it does not exist add it to prevent directory not found errors

- 

## Testing Procedure

## Things to Note

## Checklist


backend/services/implementations/imageUploadService.ts
```diff
@@ -43,6 +43,9 @@ class ImageUploadService implements IImageUploadService {
     try {
       const { createReadStream, mimetype } = await file;
       // We're using uuid() rather than file.filename to avoid conflicts of stored files in Firebase.
```
jfdoming, line 45: Nit: can you move this comment down to the line it was for

jfdoming (approved):

PR #274, opened 2022-10-25
Image upload (component + content block)

jfdoming:
## Implementation Description
- UX improvements for modals
- Extract image upload component
- Fix assorted compilation errors
- Display uploaded images
- Add size property to uploaded images
- Add loading spinner fallback
- Migrate existing image upload to new component

## Screenshots
image

- Code functionality + quality

## Testing Procedure
Manual testing

## Things to Note

## Checklist


draymond63 (dismissed):
It works, although ideally we could change a few things.
1. The modal (obv) is not exactly the same as the design. We did overlook some things in the design, so I'm cooling with changing it. We can leave at it is for this PR, but I think the description textbox should be below the image upload section. Once we add link image capabilities, I think this description textbox should appear after an image is chosen.
2. The size property is a bit simplistic, forcing everything to be a square (afaik). Could we have the size property affect the larger dimension, and allow the smaller dimension to fit the content? This would remove a lot of unnecessary whitespace and make things look a lot cleaner.

Also, check linter warnings :)

jfdoming:
> 

For 1., I agree and I'll change those in the next PR. For 2., I think that's already the behaviour? Correct me if I'm wrong, that had been my intent but if it's not working please let me know!

draymond63:
2. I uploaded `RHSLogo-horizontal.png` (from assets) which is a very wide image, and it had a lot of vertical padding. Could you give it a try? I don't think it's working as intended

PR #276, opened 2022-10-25
Module viewer mobile support

jfdoming:
## Implementation Description
- Make sidebar fill screen on smaller devices
- Adjust course padding depending on screen size
- Adjust size of videos to always fit on screen
- [unrelated bug]: Text blocks should not be editable
- Make module completion page responsive

- 

## Testing Procedure
Manual testing on smaller screen sizes (can use devtools to easily try many sizes).

## Things to Note

## Checklist

### DIFF
--- a/frontend/src/components/common/VideoPlayer.tsx
+++ b/frontend/src/components/common/VideoPlayer.tsx
@@ -1,4 +1,4 @@
-import { Center, Spinner } from "@chakra-ui/react";
+import { Center, Box, Spinner } from "@chakra-ui/react";
 import React from "react";
 import ReactPlayer from "react-player";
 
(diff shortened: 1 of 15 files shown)


draymond63 (changes requested):
I think the remain issues are:
1. Column component really squishes both things inside it. Maybe we should display its children sequentially on mobile?
2. Quiz & Matching component should reduce their padding on mobile.
3. How can you hide the sidebar without clicking on a lesson? If you are on the module completion page there doesn't seem to be a way to hide it without going back to a lesson. There should be some close button or something.

jfdoming:
> I think the remain issues are:
> 
> 1. Column component really squishes both things inside it. Maybe we should display its children sequentially on mobile?
> 2. Quiz & Matching component should reduce their padding on mobile.
> 3. How can you hide the sidebar without clicking on a lesson? If you are on the module completion page there doesn't seem to be a way to hide it without going back to a lesson. There should be some close button or something.

1. Will look into it.
2. 👍 
3. Is this a necessary feature? IMO having this sort of chevron icon in full-screen mode looks kinda weird. People can just click on the lesson they're currently in if they want to close the menu. I can add it back if you think it's necessary though.

draymond63:
3. I'd definitely prefer to have the ability, but it isn't essential. Whatever you think is best

jfdoming:
> 3. I'd definitely prefer to have the ability, but it isn't essential. Whatever you think is best

I'll spend a little time looking into whether there's a solution with a better UX

draymond63 (approved):
Looks good!

